PR #48, opened 2020-12-12
Fix AZP tests, add current_container_facts module

### LINKED ISSUE #45: Azure Pipelines vs detecting the registry endpoint
##### SUMMARY
I believe this is the only blocker to having green AZP CI on `community.docker`, though there could be some other errors which this masks.

https://dev.azure.com/ansible/community.docker/_build?view=runs

Feedback from MattClay
> It's probably due to how the tests are trying to detect the registry endpoint: https://github.com/ansible-collections/community.docker/blob/3de2cb698287ce44c99ca0599106dbd1b5519120/tests/integration/targets/setup_docker_registry/tasks/setup-frontend.yml#L70
> AZP uses user-defined docker networks while Shippable used the default bridge network.
> For example this line: https://github.com/ansible-collections/community.docker/blob/3de2cb698287ce44c99ca0599106dbd1b5519120/tests/integration/targets/setup_docker_registry/tasks/setup-frontend.yml#L70
> The `nginx_container.container.NetworkSettings.IPAddress` won't work on a user-defined network.
>  It will be found under `nginx_container.container.NetworkSettings.Networks` instead, but you have to know which network to look under.
> ansible-test required updates to its code to support user defined networks. Basically you have to determine if you're running in a container or not. If so, default network or not for the current container. Then launch other containers in the same network.
> The easiest fix for the docker tests that are affected is to only run them on real VMs. But they can work on AZP, once they're updated to support user-defined networks.
> In case it helps, here's the PR that included most of the changes to ansible-test to support user-defined docker networks: https://github.com/ansible/ansible/pull/72256

##### ISSUE TYPE
- Bug Report

felixfontein:
##### SUMMARY
Try to fix AZP tests. Fixes #45.

Includes a helper module `current_container_facts` based on https://github.com/ansible/ansible/pull/72256/files#diff-3601df6257f1aa4915de494adc6dcc84fab932ce88a6f6fec4a5efb5abebe762R69.

This also optimizes the setup_openssl role, which installes way too much for what is needed for the docker tests.

##### ISSUE TYPE
- Test Pull Request
- New Module Pull Request

##### COMPONENT NAME
current_container_facts
setup_docker
setup_docker_registry
setup_openssl

### DIFF
--- a/plugins/modules/current_container_facts.py
+++ b/plugins/modules/current_container_facts.py
@@ -1,0 +1,100 @@
+#!/usr/bin/python
+#
+# (c) 2020 Matt Clay <mclay@redhat.com>
+# (c) 2020 Felix Fontein <felix@fontein.de>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: current_container_facts
+short_description: Return facts about whether the module runs in a container
+version_added: 1.1.0
+description:
+  - Return facts about whether the module runs in a container.
+author:
+  - Felix Fontein (@felixfontein)
+'''
+
+EXAMPLES = '''
+- name: Get facts on current container
+  community.docker.current_container_facts:
+
+- name: Print information on current container when running in a container
+  ansible.builtin.debug:
+    msg: "Container ID is {{ ansible_module_container_id }}"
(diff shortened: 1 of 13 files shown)


gundalow (approved):
I don't know Docker, though this seems sensible.

Thanks for fixing this.

gundalow:
Merged, so will be in Mondays 0900 run.

felixfontein:
@gundalow thanks a lot for preparing this, and for reviewing and merging :)

PR #58, opened 2020-12-26
docker_container: return status also when 0 for non-detached containers

felixfontein:
##### SUMMARY
Resolves #26.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
docker_container

PR #59, opened 2020-12-26
docker_container: cgroup_parent

### LINKED ISSUE #6: Add support of cgroup_parent in docker_container
<!--- Verify first that your feature was not already discussed on GitHub -->
<!--- Complete *all* sections as described, this form is processed automatically -->

##### SUMMARY
<!--- Describe the new feature/improvement briefly below -->
When trying to limit container's memory using kernel_memory, memory, memory_swap on centos, it says:
```
kernel memory accounting disabled in this runc build
```
AFAIK, this is a problem with containerd.io and some people solved it using old containerd.io rpm. I trying not to downgrade packages and always stay with latest stable packages. I found the manual way using cgroup-parent feature -I'm trying to use docker-container only for change check, but it seems this module creates volumes different way that classic docker run command, which always end up as changed. So I'm stucked for now.

##### ISSUE TYPE
- Missing support for starting container with --cgroup-parent 
- Missing documentation how docker-container creates containers
- Why was kernel memory accounting disabled in containerd.io?

##### COMPONENT NAME
docker_container

##### ADDITIONAL INFORMATION
<!--- Describe how the feature would be used, why it is needed and what it would solve -->
Error from dockerd:
```
localhost dockerd[1141]: time="2020-10-07T14:07:18.456343821+02:00" level=error msg="Handler for POST /v1.40/containers/bae9b84145e6fa3b889d61910bc95655a12543eead19494c00ee2904cb145535/start returned error: OCI runtime create failed: container_linux.go:345: starting container process caused \"process_linux.go:430: container init caused \\\"process_linux.go:396: setting cgroup config for procHooks process caused \\\\\\\"kernel memory accounting disabled in this runc build\\\\\\\"\\\"\": unknown"
```
Manual limit of container
```
$ sudo dnf install libcgroup-tools
$ sudo cgcreate -g memory:influx-limit
echo 314572800  > /sys/fs/cgroup/memory/influx-limit/memory.limit_in_bytes
docker run -ti --cgroup-parent=/influx-limit/ -p 8086:8086 -v /var/lib/docker/influxdb/influxdb.conf:/etc/influxdb/influxdb.conf:ro -v /mnt/influxdb/:/var/lib/influxdb:rw "soc/influx:alpine" 
```
Create using ansible:
```yaml
# tasks:
- name: Influx - Check container changes
  docker_container:
    name: influxdb
    image: "{{ harbor_registry }}/soc/influx:alpine"
    restart_policy: always
    ports: "{{ influxdb_port }}:8086"
    volumes:
      - "{{ influxdb_dir }}/influxdb.conf:/etc/influxdb/influxdb.conf:ro"
      - "{{ influxdb_datadir }}:/var/lib/influxdb:rw"
    state: started
  check_mode: yes
  diff: yes
  notify: Recreate influxdb
  tags: influx

- name: Influx - Create cgroup limit group
  command: cgcreate -g memory:influxdb-limit
  register: output
  changed_when: "output.stdout != ''"

- name: Influx - Setup influxdb cgroup limit
  shell: "echo {{ influxdb_cgroup_parent.memory }} > /sys/fs/cgroup/memory/influxdb-limit/memory.limit_in_bytes"
  notify: Recreate influxdb

# handler:
- name: Recreate influxdb
  shell: >
    docker ps | grep -c influxdb && docker stop influxdb && docker rm influxdb;   
    docker run -d
    --name influxdb
    --init=false
    -p 0.0.0.0:{{ influxdb_port }}:8086
    -v {{ influxdb_dir }}/influxdb.conf:/etc/influxdb/influxdb.conf:ro
    -v {{ influxdb_datadir }}:/var/lib/influxdb:rw
    --restart=always
    --cgroup-parent=/influxdb-limit/
    {{ harbor_registry }}/soc/influx:alpine
```
but there is always change:
```
TASK [influx : Influx - Check container] ********************************************************************************************************************
Wednesday 07 October 2020  12:45:11 +0200 (0:00:01.136)       0:00:06.984 *****                                                                              
--- before                                                                                                                                                   
+++ after                                                                                                                                                    
@@ -1,5 +1,6 @@                                                                                                                                               
{                                                                                                                                                                
"expected_volumes": [                                                                                                                                   +        "/etc/influxdb/influxdb.conf",                                                                                                                      
         "/var/lib/influxdb"                                                                                                                                 
     ]                                                                                                                                                       
 }                                                                                                                                                           
                                                                                                                                                             
changed: [vm]   
```
I also tried --mount type=bind ..., but result was the same.

When looking on docker inspect, both commands, ansible's docker_container as well as docker run create bind mounts as they should, but docker_container also changes container's config, and that is the cause of that change I believe. But I did not find a way how to adjust it, it seems the docker generates this config file when starting container.

So, the easiest would be to add cgroup-parent into docker_container such as podman has, IMHO https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html#parameter-cgroup_parent

Thank you. 
<!--- HINT: You can also paste gist.github.com links for larger files -->


```
$ ansible --version
ansible 2.9.9
  config file = /home/user/SOC/brickleberry/ansible/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/.local/lib/python3.7/site-packages/ansible
  executable location = /home/user/.local/bin/ansible
  python version = 3.7.9 (default, Aug 19 2020, 17:05:11) [GCC 9.3.1 20200408 (Red Hat 9.3.1-2)]
```

felixfontein:
##### SUMMARY
Fixes #6.

Unfortunately cgroup_parent is pretty much undocumented, the docs do not seem to mention it at all. The feature was added in Docker SDK for Python 1.4.0 (https://github.com/docker/docker-py/commit/3caaa0050b044462157b755f3810e44857d73761), so we can use it with all supported versions of the Docker SKD for Python (we require >= 1.8.0). It has already been mentioned in the API docs for 1.19 (https://docs.docker.com/engine/api/v1.19/#operation/ContainerCreate), so it should be available for all Docker API versions we support (>= 1.20).

This PR also mentions in all Docker SDK for Python based modules that they use the Docker SDK for Python to communicate with the docker daemon, since that is apparently unclear according to #6 ("Missing documentation how docker-container creates containers").

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
docker_container

### DIFF
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -52,6 +52,11 @@ options:
       - List of capabilities to drop from the container.
     type: list
     elements: str
+  cgroup_parent:
+    description:
+      - Specify the parent cgroup for the container.
+    type: str
+    version_added: 1.1.0
   cleanup:
     description:
       - Use with I(detach=false) to remove the container after successful execution.
@@ -1589,6 +1594,7 @@ class TaskParameters(DockerBaseClass):
             publish_all_ports='publish_all_ports',
             links='links',
             privileged='privileged',
+            cgroup_parent='cgroup_parent',
             dns='dns_servers',
             dns_opt='dns_opts',
             dns_search='dns_search_domains',
@@ -2175,6 +2181,7 @@ class Container(DockerBaseClass):
             interactive=config.get('OpenStdin'),
             capabilities=host_config.get('CapAdd'),
             cap_drop=host_config.get('CapDrop'),
+            cgroup_parent=host_config.get('CgroupParent'),
             expected_devices=host_config.get('Devices'),
             dns_servers=host_config.get('Dns'),
             dns_opts=host_config.get('DnsOptions'),
@@ -3371,6 +3378,7 @@ def main():
         blkio_weight=dict(type='int'),
         capabilities=dict(type='list', elements='str'),
         cap_drop=dict(type='list', elements='str'),
+        cgroup_parent=dict(type='str'),
         cleanup=dict(type='bool', default=False),
         command=dict(type='raw'),
         comparisons=dict(type='dict'),
--- a/changelogs/fragments/59-docker_container-cgroup-parent.yml
+++ b/changelogs/fragments/59-docker_container-cgroup-parent.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- "docker_container - support specifying ``cgroup_parent`` (https://github.com/ansible-collections/community.docker/issues/6, https://github.com/ansible-collections/community.docker/pull/59)."
--- a/plugins/doc_fragments/docker.py
+++ b/plugins/doc_fragments/docker.py
@@ -111,6 +111,9 @@ notes:
 
     DOCKER_PY_1_DOCUMENTATION = r'''
 options: {}
+notes:
+  - This module uses the L(Docker SDK for Python,https://docker-py.readthedocs.io/en/stable/) to
+    communicate with the Docker daemon.
 requirements:
   - "Docker SDK for Python: Please note that the L(docker-py,https://pypi.org/project/docker-py/)
      Python module has been superseded by L(docker,https://pypi.org/project/docker/)
@@ -127,6 +130,9 @@ requirements:
 
     DOCKER_PY_2_DOCUMENTATION = r'''
 options: {}
+notes:
+  - This module uses the L(Docker SDK for Python,https://docker-py.readthedocs.io/en/stable/) to
+    communicate with the Docker daemon.
 requirements:
   - "Python >= 2.7"
   - "Docker SDK for Python: Please note that the L(docker-py,https://pypi.org/project/docker-py/)
--- a/tests/integration/targets/docker_container/tasks/tests/options.yml
+++ b/tests/integration/targets/docker_container/tasks/tests/options.yml
@@ -158,6 +158,40 @@
     - capabilities_4 is changed
 
 ####################################################################
+## cgroup_parent ###################################################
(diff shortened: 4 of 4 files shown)


felixfontein:
@WojciechowskiPiotr if you have time, can you take a quick look at this PR and #58?

plugins/modules/docker_container.py
```diff
@@ -52,6 +52,11 @@
       - List of capabilities to drop from the container.
     type: list
     elements: str
+  cgroup_parent:
+    description:
+      - Specify the parent cgroup for the container.
```
bmillemathias, line 57: I see in the tests you added alongside you pass an empty string, what does it do? perhaps it should be added this is supported and the behaviour ?

felixfontein: Well, the test doesn't really do anything :) The problem with testing this properly in integration tests is that you need to know which cgroups are around on the systems you run the tests on, which is pretty much impossible since they can run on any system (if you run `ansible-test integration --docker xxx` locally, your own docker daemon will be used, and thus it depends on your local system). That's why I put in an empty string, which by the module is treated as "value specified by user", but by the Docker daemon as "no value supplied" resp. "user does not care whether/which cgroup parent to use". This test mainly makes sure that all related code is run through at least once.

tadeboro (approved):

felixfontein:
@bmillemathias @tadeboro thanks a lot for reviewing this!

PR #60, opened 2020-12-26
docker_container: deprecate current behavior when specifying all with other port mappings

felixfontein:
##### SUMMARY
Regarding #8, deprecate the current behavior when `published_ports` contains port mappings next to the special value `all`.

The deprecation meesage informs user that the current behavior (which they might not be aware of) will change in a breaking way, namely either it will be forbidden, or properly handled (like Docker CLI) in community.docker 2.0.0.

I didn't want to promise that I implement it properly, that's why I keep the possibliity to simply forbid it. In that case, specifying both `all` and port mappings will make the module fail. This can be changed anytime later (would be a new feature at that point).

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
docker_container


ssbarnea (approved):

felixfontein:
@ssbarnea thanks for reviewing this!

PR #61, opened 2020-12-26
Docker inventory plugin

felixfontein:
##### SUMMARY
Adds a docker inventory plugin for adding containers to the inventory.

The main incentive is to allow people to stop using the inventory script (which isn't even part of this repo, as I noticed today: https://github.com/ansible-collections/community.general/blob/main/scripts/inventory/docker.py).

##### ISSUE TYPE
- New Plugin Pull Request

##### COMPONENT NAME
docker


felixfontein:
(The 'failed' code coverage test is because not all tests were run. The new code is mostly covered, what's missing there are mainly error situations (listing containers fails etc.).

plugins/inventory/docker.py
```diff
@@ -0,0 +1,321 @@
...
+            - "C(stopped): contains all containers that are not running."
+            - If this is not set to C(true), you should use keyed groups to add the containers to groups.
+              See the examples for how to do that.
+        type: bool
+        default: false
+'''
+
+EXAMPLES = '''
+# Minimal example using local Docker daemon
+plugin: community.docker.docker
```
WojciechowskiPiotr, line 100: I think the `docker` is a misleading name. This inventory plugin is for containers so in my opinion it should be named as `docker_container` or `docker_containers` which will match the naming convention used for the other two plugins.

plugins/inventory/docker.py
```diff
@@ -0,0 +1,321 @@
...
+        default: 127.0.0.1
+
+    private_ssh_port:
+        description:
+            - The port containers use for SSH.
+            - Only used if I(connection_type) is C(ssh).
+        type: int
+        default: 22
+
+    add_legacy_groups:
```
WojciechowskiPiotr, line 79: Why name it as a `legacy groups`? Why not `group_by` giving users a choice if they want to use particular groups. There we may add `legacy` as an option that will return output like the old script?

felixfontein: Users can already create as many groups as they want in a very flexible way with `groups` and `keyed_groups`. I don't see why to add more choice. This here is mainly a convenience option to ease migration from the old inventory script.

plugins/inventory/docker.py
felixfontein: Good idea! I kept the name similar to the old inventory plugin, but there's really no need for that. I think `docker_containers` sounds good. Actually, we could also call it just `containers`, since the FQCN `community.docker.containers` would already include "docker".

felixfontein: I've now named it `docker_containers` for consistency.

WojciechowskiPiotr (approved):

felixfontein:
@WojciechowskiPiotr thanks a lot for reviewing this!

PR #66, opened 2021-01-07
Fix IPv6 zone handling

### LINKED ISSUE #65: docker_container: IPv6 port validation doesn't allow for interface specification.
https://github.com/ansible-collections/community.docker/blob/c7a3d9f057f853c99028430f123a26effe433b03/plugins/modules/docker_container.py#L1732

My link local address is of the format `[fe80::d65d:daaf:fe1e:11a7%eno1]`, and I'm trying to use that in a port map as follows:

```
        ports:
          - "[fe80::d65d:dfff:fe1e:11a8%eno1]:53:53/tcp"
```

But the regex doesn't accept this, even though it's a valid bind.

Manually creating the container with that port mapping does work.

felixfontein:
##### SUMMARY
Fixes #65. The same problem also appears in docker_network's CIDR validation. I don't know whether Docker supports zones here, but I don't see why we should restrict to non-zone CIDRs.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
docker_container
docker_network

### DIFF
--- a/changelogs/fragments/66-ipv6-zones.yml
+++ b/changelogs/fragments/66-ipv6-zones.yml
@@ -1,0 +1,3 @@
+bugfixes:
+- "docker_container - allow IPv6 zones (RFC 4007) in bind IPs (https://github.com/ansible-collections/community.docker/pull/66)."
+- "docker_image - allow IPv6 zones (RFC 4007) in network CIDRs (https://github.com/ansible-collections/community.docker/pull/66)."
--- a/changelogs/fragments/66-ipv6-zones.yml
+++ b/changelogs/fragments/66-ipv6-zones.yml
@@ -1,3 +1,2 @@
 bugfixes:
 - "docker_container - allow IPv6 zones (RFC 4007) in bind IPs (https://github.com/ansible-collections/community.docker/pull/66)."
-- "docker_image - allow IPv6 zones (RFC 4007) in network CIDRs (https://github.com/ansible-collections/community.docker/pull/66)."
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -1756,7 +1756,7 @@ class TaskParameters(DockerBaseClass):
             elif p_len == 3:
                 # We only allow IPv4 and IPv6 addresses for the bind address
                 ipaddr = parts[0]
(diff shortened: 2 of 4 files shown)


WojciechowskiPiotr:
Is there any way we can test IPv6 zones in integration tests?

felixfontein:
@WojciechowskiPiotr probably not in a sane way. I'm not sure which values are actually valid for it, and it probably depends on the system running the docker daemon, which at least in the dockerized tests we have no access to.

WojciechowskiPiotr:
> @WojciechowskiPiotr probably not in a sane way. I'm not sure which values are actually valid for it, and it probably depends on the system running the docker daemon, which at least in the dockerized tests we have no access to.

Did you manage to manually test the patch?

felixfontein:
No. I only tested the regex itself. I'm hoping that @Kristof-Mattei could try it out.

WojciechowskiPiotr:
> No. I only tested the regex itself. I'm hoping that @Kristof-Mattei could try it out.

Patch looks fine, but let's test properly it before merging

felixfontein:
@Kristof-Mattei you can use the way described here: https://docs.ansible.com/ansible/devel/user_guide/collections_using.html#installing-a-collection-from-a-git-repository Here, it would be`ansible-galaxy collection install git+https://github.com/felixfontein/community.docker.git,ipv6-zones`. Just make sure to eventually delete the manually installed collection, because otherwise it will hide updated versions of the collections installed with Ansible (if you are using Ansible 2.10).

Alternatively (if you don't mind working with `git`) you could use the way described here (https://docs.ansible.com/ansible/devel/dev_guide/developing_collections.html#contributing-to-collections, third paragraph+). You need to check out the `ipv6-zones` branch of https://github.com/felixfontein/community.docker/.

felixfontein:
@Kristof-Mattei did you had a chance to test this yet?

felixfontein:
I managed to test it for docker_container, it seems to work fine. I couldn't test it for docker_network (resp. Docker daemon always complained about an invalid CIDR), so I removed the change for that one.

felixfontein:
As Docker daemon seems to allow a random zone, we're able to (somewhat) test this in CI.

Andersson007 (approved):
LGTM

felixfontein:
@WojciechowskiPiotr @Andersson007 thanks for reviewing this!

PR #84, opened 2021-02-10
Update CI

felixfontein:
##### SUMMARY
See https://github.com/ansible-collections/overview/issues/45#issuecomment-776859063 and the change below.

I made sure that the targets removed from devel are covered by stable-2.10.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
There's an effective increase of two targets.

PR #87, opened 2021-02-17
docker_image: improve/fix handling of image IDs

felixfontein:
##### SUMMARY
This PR allows to use image IDs in more contexts in docker_image. It also makes the code for loading Docker images aware of image IDs, and fixes that part of #86.

It does *not* allow to load multiple images though. That should be done in a new module.

##### ISSUE TYPE
- Feature Pull Request
- Bugfix Pull Request

##### COMPONENT NAME
docker_image

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -534,6 +534,9 @@ class AnsibleDockerClientBase(Client):
         if len(images) == 1:
             try:
                 inspection = self.inspect_image(images[0]['Id'])
+            except NotFound:
+                self.log("Image %s:%s not found." % (name, tag))
+                return None
             except Exception as exc:
                 self.fail("Error inspecting image %s:%s - %s" % (name, tag, str(exc)))
             return inspection
@@ -541,7 +544,7 @@ class AnsibleDockerClientBase(Client):
         self.log("Image %s:%s not found." % (name, tag))
         return None
 
-    def find_image_by_id(self, image_id):
+    def find_image_by_id(self, image_id, accept_not_there=False):
         '''
         Lookup an image (by ID) and return the inspection results.
         '''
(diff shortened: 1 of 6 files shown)


felixfontein:
ready_for_review

felixfontein:
While working on #90, I found a couple of bugs / oversights that are fixed in 5248b4112d28506929d1b23f31bfa6472581cead.

felixfontein:
Ok, this is now ready.

aminvakil (approved):
LGTM, but I would wait for someone else's review with more experience in python instead of myself :)

plugins/module_utils/common.py
```diff
@@ -551,6 +554,11 @@ def find_image_by_id(self, image_id):
         self.log("Find image %s (by ID)" % image_id)
         try:
             inspection = self.inspect_image(image_id)
+        except NotFound as exc:
+            if not accept_not_there:
```
russoz, line 558: I always try to avoid variable names with `not` in them. This line illustrates a bit why: reading it aloud, it goes like "if not accept not there". Maybe `accept_missing_image` ?

plugins/modules/docker_image.py
```diff
@@ -463,6 +480,9 @@ def absent(self):
             if not self.check_mode:
                 try:
                     self.client.remove_image(name, force=self.force_absent)
+                except NotFound as dummy:
```
russoz, line 483: Since we are not really using it, I'd remove the `as dummy` part altogether. It does not make the code better at all.

russoz:
LGTM other than the couple of minor issues above, but also putting a disclaimer: I have not used the docker modules, so not very confident on what impacts these changes might have.

plugins/module_utils/common.py
felixfontein: Fixed in 7f0ec90ac5b01f09bed827fece52e52e4f754c9e.

plugins/modules/docker_image.py
felixfontein: Fixed in b3b6da6165168176d99abfc0d7d0e8ee3a18cfb5.

felixfontein:
@aminvakil @russoz thanks a lot for reviewing this!

PR #89, opened 2021-02-18
docker_image: allow to specify pull platform

### LINKED ISSUE #79: Add 'platform' option when pulling in docker_image
##### SUMMARY

It would be handy it was possible to specify platform when pulling an image. Currently when pulling multiarch image that doesn't have a special tag for each architecture it automatically starts pulling the image with architecture of userspace and there's no way to override this behaviour other than using cli:

```yaml
- command:  docker pull --platform=linux/arm64 minio/minio:RELEASE.2021-01-16T02-19-44Z
```

- i.e. when on Raspbian with aarch64 kernel and armv7 userspace, it pulls armv7 image and fails if it doesn't exist, manual override using `--platform arm64` is needed to fix it, but `docker_image` doesn't support `platform` option when pulling

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
docker_image

##### ADDITIONAL INFORMATION

- Related issue: https://github.com/ansible-collections/community.docker/issues/22

felixfontein:
##### SUMMARY
Fixes #79.
##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
docker_image

### DIFF
--- a/plugins/modules/docker_image.py
+++ b/plugins/modules/docker_image.py
@@ -163,6 +163,8 @@ options:
   pull_platform:
     description:
       - When pulling an image, ask for this specific platform.
+      - Note that this value is not used to determine whether the image needs to be pulled. This might change
+        in the future in a minor release, though.
     type: str
     version_added: 1.3.0
   push:
--- a/plugins/modules/docker_image.py
+++ b/plugins/modules/docker_image.py
@@ -160,13 +160,18 @@ options:
       - Note that image IDs (hashes) are not supported.
     type: str
     required: yes
-  pull_platform:
+  pull:
     description:
-      - When pulling an image, ask for this specific platform.
-      - Note that this value is not used to determine whether the image needs to be pulled. This might change
-        in the future in a minor release, though.
-    type: str
+      - "Specifies options used for pulling images."
+    type: dict
     version_added: 1.3.0
+    suboptions:
+      platform:
+        description:
+          - When pulling an image, ask for this specific platform.
+          - Note that this value is not used to determine whether the image needs to be pulled. This might change
+            in the future in a minor release, though.
+        type: str
   push:
     description:
       - Push the image to the registry. Specify the registry as part of the I(name) or I(repository) parameter.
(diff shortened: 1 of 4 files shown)


felixfontein:
ready_for_review

plugins/modules/docker_image.py
```diff
@@ -818,6 +827,7 @@ def detect_platform(client):
     option_minimal_versions["build.use_config_proxy"] = dict(docker_py_version='3.7.0', detect_usage=detect_use_config_proxy)
     option_minimal_versions["build.etc_hosts"] = dict(docker_py_version='2.6.0', docker_api_version='1.27', detect_usage=detect_etc_hosts)
     option_minimal_versions["build.platform"] = dict(docker_py_version='3.0.0', docker_api_version='1.32', detect_usage=detect_platform)
+    option_minimal_versions["pull_platform"] = dict(docker_py_version='3.0.0')
```
WojciechowskiPiotr, line 830: I think this feature requires API at least 1.32 (https://docs.docker.com/engine/reference/commandline/pull/)

plugins/modules/docker_image.py
```diff
@@ -160,6 +160,13 @@
       - Note that image IDs (hashes) are not supported.
     type: str
     required: yes
+  pull_platform:
```
WojciechowskiPiotr, line 163: We have `platform` as a suboption in `build`. I know at least now there are no suboptions for `pull` but maybe for consistency, this should be a `platform` as suboption in `pull` not as `pull_platform` option?

plugins/modules/docker_image.py
felixfontein: Yep, that's also what docker-py is saying. Will change that in a minute.

plugins/modules/docker_image.py
felixfontein: Hmm, that would definitely make it easier to extend in the future. Let me try this out.

felixfontein: I tried that in 2737cf4f15dfad5423e407239058a61d58035d0c. I also used `apply_defaults=True` for `build`, that should simplify the code. Let's see what CI says :)

felixfontein: Ok, the failing tests reminded me why `apply_defaults=True` is not a good idea: it also enforces required fields, which is something we do not want here :)

felixfontein:
@WojciechowskiPiotr should be ready for review again :)

felixfontein:
@WojciechowskiPiotr thanks for reviewing and merging! :)

WojciechowskiPiotr:
> @WojciechowskiPiotr should be ready for review again :)

Did it while checks were running :)

Issue #7, opened 2020-10-02
docker_swarm_service does not expose --replicas-max-per-node


nicholasamorim:
##### SUMMARY
docker_swarm_service does not expose the `--replicas-max-per-node` that is very useful especially when running services that might need HA in a more strict manner (databases, etc)

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

docker_swarm_service

##### ADDITIONAL INFORMATION

```yaml

```

b2az:
just here to tell you, that i also need this in my playbooks.

WojciechowskiPiotr:
> just here to tell you, that i also need this in my playbooks.

Hi, the docker-py maintainers review my change proposal, so I how we can marge it soon. I have the patch for docker_swarm_service ready.

WojciechowskiPiotr:
My change (https://github.com/docker/docker-py/pull/2741) has been merged so as soon as the next version of docker-py is released I will implement this feature in `docker_swarm_service`

WojciechowskiPiotr:
FYI my change in docker-py will be included in release 4.5.0 so we have to wait until it is released.

hartwork:
> FYI my change in docker-py will be included in release 4.5.0 so we have to wait until it is released.

While [not mentioned in the change log of docker-py 4.4.3](https://github.com/docker/docker-py/issues/2774) I can confirm that https://github.com/docker/docker-py/issues/2774 has just hit PyPI in release 4.4.3 already yesterday.

felixfontein:
As long as they don't remove it again in 4.4.4, I guess we can start using it ;-)

WojciechowskiPiotr:
Ok, I saw it on the list for the 4.5.0 release but it is even better it was included in 4.4.3 even by this small mistake :) In such case, I will prepare my change for community.docker shortly.

felixfontein:
This has been implemented in #92 and will be available in the 1.3.0 release.

Issue #91, opened 2021-02-23
Feature Request: community.docker.docker_container --storage-opt


mujocloud:
SUMMARY
Hi Team,

I would like to request a new feature addition to the docker_container module -
When we run a container there is an option to specify the size of the container's root volume docker run -itd --storage-opt
Please implement that in the docker_container module as well

ISSUE TYPE
Feature Idea
COMPONENT NAME
community.docker.docker_container module

ADDITIONAL INFORMATION
If the team could add this feature it would allow all docker containers running on XFS to have a pre-set amount of storage and not use up the entire system's storage
https://fabianlee.org/2020/01/15/docker-use-overlay2-with-an-xfs-backing-filesystem-to-limit-rootfs-size/
```
   tasks:
      - name: run container with rootfs size limit
        docker_container:
            name: "{{ container_name }}"
            image: "{{ image_name }}"
            detach: yes
            tty: yes
            state: started
           storage-opt: 
              - size: <size in GB>
           ports:
               - "{{ port }}"
```

felixfontein:
`create_host_config` (https://docker-py.readthedocs.io/en/stable/api.html#docker.api.container.ContainerApiMixin.create_host_config) has an option `storage_opt` of type `dict`, so this would be a dictionary and not a list.

felixfontein:
resolved_by_pr #93

mujocloud:
Awesome!! Thanks a lot!!

PR #90, opened 2021-02-18
Add docker_image_load module

felixfontein:
##### SUMMARY
First basic, untested version. Fixes #86.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
docker_image_load

PR #92, opened 2021-02-23
Replicas max per node

WojciechowskiPiotr:
##### SUMMARY
Adding support for MaxReplicas Placement parameter for `docker_swarm_service` (requested in https://github.com/ansible-collections/community.docker/issues/7)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
docker_swarm_service

### DIFF
--- a/changelogs/fragments/92-replicas-max-per-node.yml
+++ b/changelogs/fragments/92-replicas-max-per-node.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - "docker_swarm_service - Adding support for maximum number of tasks per node (replicas_max_per_node) when running swarm service in replicated mode. Introduced in API 1.40"
+  - "docker_swarm_service - adding support for maximum number of tasks per node (``replicas_max_per_node``) when running swarm service in replicated mode. Introduced in API 1.40 (https://github.com/ansible-collections/community.docker/issues/7, https://github.com/ansible-collections/community.docker/pull/92)."
--- a/changelogs/fragments/92-replicas-max-per-node.yml
+++ b/changelogs/fragments/92-replicas-max-per-node.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "docker_swarm_service - Adding support for maximum number of tasks per node (replicas_max_per_node) when running swarm service in replicated mode. Introduced in API 1.40"
--- a/plugins/modules/docker_swarm_service.py
+++ b/plugins/modules/docker_swarm_service.py
@@ -13,6 +13,7 @@ author:
   - "Dario Zanzico (@dariko)"
   - "Jason Witkowski (@jwitko)"
   - "Hannes Ljungberg (@hannseman)"
+  - "Piotr Wojciechowski (@wojciechowskipiotr)"
 short_description: docker swarm service
 description:
   - Manages docker services via a swarm manager node.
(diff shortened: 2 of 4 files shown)


plugins/modules/docker_swarm_service.py
```diff
@@ -344,6 +345,12 @@
           - Requires API version >= 1.27.
         type: list
         elements: dict
+      replicas_max_per_node:
+        description:
+          - Maximum number of tasks per node.
+          - Corresponds to the C(--replicas_max_per_node) option of C(docker service create).
+          - Requires API version >= 1.40
+        type: int
```
felixfontein, line 353:
```suggestion
        type: int
        version_added: 1.3.0
```

felixfontein (commented):
I think you also need to update the return value documentation, since you added a new field. (I think that new field is also what makes the tests fail.)

plugins/modules/docker_swarm_service.py
WojciechowskiPiotr: Let's see if this is the problem. Iinteresting that only tests on two versions of RHEL failed.

felixfontein: The tests probably only run on that two nodes. (All swarm-related tests don't run in the docker containers, only on VMs.)

WojciechowskiPiotr:
ready_for_review

felixfontein (commented):
You should also mention `replicas_max_per_node` in the return example (https://github.com/ansible-collections/community.docker/blob/8f209e4070f82c4d0f9c0b876a952b5fa4d0ad17/plugins/modules/docker_swarm_service.py#L728). Also this needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

Besides that, looks good to me. As far as I can judge changes to this module :)

@hannseman @jwitko @dariko any comments on this?

hannseman (commented):
Looks good to me! 👍

WojciechowskiPiotr:
> You should also mention `replicas_max_per_node` in the return example (
> 
> https://github.com/ansible-collections/community.docker/blob/8f209e4070f82c4d0f9c0b876a952b5fa4d0ad17/plugins/modules/docker_swarm_service.py#L728
> 
> ). Also this needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).
> Besides that, looks good to me. As far as I can judge changes to this module :)
> 
> @hannseman @jwitko @dariko any comments on this?

Updated

changelogs/fragments/92-replicas-max-per-node.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "docker_swarm_service - Adding support for maximum number of tasks per node (replicas_max_per_node) when running swarm service in replicated mode. Introduced in API 1.40"
```
felixfontein, line 2:
```suggestion
  - "docker_swarm_service - adding support for maximum number of tasks per node (``replicas_max_per_node``) when running swarm service in replicated mode. Introduced in API 1.40 (https://github.com/ansible-collections/community.docker/issues/7, https://github.com/ansible-collections/community.docker/pull/92)."
```

felixfontein (approved):

felixfontein:
@WojciechowskiPiotr thanks for implementing this!
@hannseman thanks for reviewing!

If you have some time, can you take a look at #87, #89, #90? :) I'll also try to create a PR for #91 today, I think with that we can release 1.3.0 very soon.

PR #96, opened 2021-03-01
Windows Container support

tmeckel:
##### SUMMARY

This PR contains changes to `docker.py` to support  Windows Containers.

##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/changelogs/fragments/96-windows-support.yaml
+++ b/changelogs/fragments/96-windows-support.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - windows container support - added support for Windows Containers
+  - docker_container - added support for Windows containers (https://github.com/ansible-collections/community.docker/pull/96).
--- a/changelogs/fragments/96-windows-support.yaml
+++ b/changelogs/fragments/96-windows-support.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - windows container support - added support for Windows Containers
--- a/changelogs/fragments/96-windows-support.yaml
+++ b/changelogs/fragments/96-windows-support.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - docker_container - added support for Windows containers (https://github.com/ansible-collections/community.docker/pull/96).
+  - docker connection plugin - added support for Windows containers (https://github.com/ansible-collections/community.docker/pull/96).
--- a/plugins/connection/docker.py
+++ b/plugins/connection/docker.py
@@ -280,10 +280,6 @@ class Connection(ConnectionBase):
         if getattr(self._shell, "_IS_WINDOWS", False) and stderr.startswith(b"#< CLIXML"):
             stderr = _parse_clixml(stderr)
 
-        display.debug(u"p.returncode == {0}".format(p.returncode))
-        display.debug(u"stderr == {0}".format(stderr))
-        display.debug(u"stdout == {0}".format(stdout))
-
         return (p.returncode, stdout, stderr)
 
     def _prefix_login_path(self, remote_path):
@@ -357,18 +353,14 @@ $strm.Dispose() }''' % (out_path)
             args = [to_bytes(i, errors='surrogate_or_strict') for i in args]
             try:
                 p = subprocess.Popen(
-                        args,
-                        stdin=in_file,
-                        stdout=subprocess.PIPE,
-                        stderr=subprocess.PIPE
-                    )
+                    args,
+                    stdin=in_file,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE
+                )
             except OSError:
                 raise AnsibleError("docker connection requires dd command in the container to put files")
             stdout, stderr = p.communicate()
-
-            display.debug(u"p.returncode == {0}".format(p.returncode))
-            display.debug(u"stderr == {0}".format(stderr))
-            display.debug(u"stdout == {0}".format(stdout))
 
             if p.returncode != 0:
                 raise AnsibleError("failed to transfer file %s to %s:\n%s\n%s" %
--- a/plugins/connection/docker.py
+++ b/plugins/connection/docker.py
@@ -57,7 +57,6 @@ from ansible.module_utils.six.moves import shlex_quote
 from ansible.module_utils._text import to_bytes, to_native, to_text
 from ansible.plugins.connection import ConnectionBase, BUFSIZE
 from ansible.utils.display import Display
-from ansible.plugins.shell.powershell import _parse_clixml
 
 display = Display()
 
@@ -278,6 +277,7 @@ class Connection(ConnectionBase):
 
         # When running on Windows, stderr may contain CLIXML encoded output
         if getattr(self._shell, "_IS_WINDOWS", False) and stderr.startswith(b"#< CLIXML"):
+            from ansible.plugins.shell.powershell import _parse_clixml
             stderr = _parse_clixml(stderr)
 
         return (p.returncode, stdout, stderr)
@@ -346,9 +346,7 @@ $strm.Dispose() }''' % (out_path)
                 args = self._build_exec_cmd(["cmd", "/c", self._shell._encode_script(cmd, as_list=False, strict_mode=False, preserve_rc=False)])
                 display.debug(u"Command to execute: {0}".format(args))
             else:
-                out_path = self._prefix_login_path(out_path)
-                out_path = shlex_quote(out_path)
-            args = self._build_exec_cmd([self._play_context.executable, "-c", "dd of=%s bs=%s%s" % (out_path, BUFSIZE, count)])
+                args = self._build_exec_cmd([self._play_context.executable, "-c", "dd of=%s bs=%s%s" % (out_path, BUFSIZE, count)])
 
             args = [to_bytes(i, errors='surrogate_or_strict') for i in args]
             try:
--- a/plugins/connection/docker.py
+++ b/plugins/connection/docker.py
@@ -13,7 +13,7 @@ DOCUMENTATION = '''
     author:
         - Lorin Hochestein (!UNKNOWN)
         - Leendert Brouwer (!UNKNOWN)
-    name: docker
+    connection: docker
     short_description: Run tasks in docker containers
     description:
         - Run commands or put/fetch files to an existing docker container.
@@ -57,6 +57,7 @@ from ansible.module_utils.six.moves import shlex_quote
 from ansible.module_utils._text import to_bytes, to_native, to_text
 from ansible.plugins.connection import ConnectionBase, BUFSIZE
 from ansible.utils.display import Display
+from ansible.plugins.shell.powershell import _parse_clixml
 
 display = Display()
 
@@ -64,7 +65,7 @@ display = Display()
 class Connection(ConnectionBase):
     ''' Local docker based connections '''
 
-    transport = 'community.docker.docker'
+    transport = 'community.general.docker'
     has_pipelining = True
 
     def __init__(self, play_context, new_stdin, *args, **kwargs):
@@ -81,7 +82,9 @@ class Connection(ConnectionBase):
 
         # Windows uses Powershell modules
         if getattr(self._shell, "_IS_WINDOWS", False):
+            self.has_native_async = True
             self.module_implementation_preferences = ('.ps1', '.exe', '')
+            self.allow_executable = False
 
         if 'docker_command' in kwargs:
             self.docker_cmd = kwargs['docker_command']
@@ -214,7 +217,11 @@ class Connection(ConnectionBase):
         """ Run a command on the docker host """
         super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
 
-        local_cmd = self._build_exec_cmd([self._play_context.executable, '-c', cmd])
+        if getattr(self._shell, "_IS_WINDOWS", False):
+            sudoable = False
+            local_cmd = self._build_exec_cmd(["cmd", "/c", cmd])
+        else:
+            local_cmd = self._build_exec_cmd([self._play_context.executable, '-c', cmd])
 
         display.vvv(u"EXEC {0}".format(to_text(local_cmd)), host=self._play_context.remote_addr)
         display.debug("opening command with Popen()")
@@ -268,6 +275,15 @@ class Connection(ConnectionBase):
         display.debug("done communicating")
 
         display.debug("done with docker.exec_command()")
+
+        # When running on Windows, stderr may contain CLIXML encoded output
+        if getattr(self._shell, "_IS_WINDOWS", False) and stderr.startswith(b"#< CLIXML"):
+            stderr = _parse_clixml(stderr)
+
+        display.debug(u"p.returncode == {0}".format(p.returncode))
+        display.debug(u"stderr == {0}".format(stderr))
+        display.debug(u"stdout == {0}".format(stdout))
+
         return (p.returncode, stdout, stderr)
 
     def _prefix_login_path(self, remote_path):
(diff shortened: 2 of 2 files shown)


felixfontein (commented):
Thanks for this contribution!

I can't review this fully since I don't have or use Windows; I'll mainly make sure that the code looks good and that it won't break the non-Windows side :-)

Two things:
1. This needs a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to);
2. Please remove the `display.debug` statements that dump stdout/stderr. This could accidentally dump secrets into logs where they do not belong.

tmeckel:
@felixfontein Thank for the quick reply on my PR! 
I changed the code and added a changelog fragment.

changelogs/fragments/96-windows-support.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - windows container support - added support for Windows Containers
```
felixfontein, line 2:
```suggestion
  - docker_container - added support for Windows containers (https://github.com/ansible-collections/community.docker/pull/96).
```

plugins/connection/docker.py
```diff
@@ -57,6 +57,7 @@
 from ansible.module_utils._text import to_bytes, to_native, to_text
 from ansible.plugins.connection import ConnectionBase, BUFSIZE
 from ansible.utils.display import Display
+from ansible.plugins.shell.powershell import _parse_clixml
```
felixfontein, line 60: I think it would be better to wrap that import in try/except, similarly to how it is described here: https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/dev_guide/testing/sanity/import.rst

I'd prefer that if this import ever breaks (for whatever reason, could also be `xml` not being available that's imported by the `powershell` shell plugin, or the shell plugin being moved to a collection that's not a dependency of `community.docker`), this won't break the docker connection plugin for everyone (but only for Windows containers).

plugins/connection/docker.py
```diff
@@ -308,11 +330,34 @@ def put_file(self, in_path, out_path):
...
+    }
+} while ($read -gt 0)
+$fstrm.Flush()
+$fstrm.Dispose()
+$strm.Dispose() }''' % (out_path)
+                args = self._build_exec_cmd(["cmd", "/c", self._shell._encode_script(cmd, as_list=False, strict_mode=False, preserve_rc=False)])
+                display.debug(u"Command to execute: {0}".format(args))
+            else:
+                out_path = self._prefix_login_path(out_path)
+                out_path = shlex_quote(out_path)
```
felixfontein, line 350: I think these two lines should better be in another PR (with changelog fragment) - this looks like a bugfix (that should be backported), not like a new feature (like Windows container support, that should not be backported).

tmeckel: I removed those two lines

plugins/connection/docker.py
tmeckel: I moved the import right at the position where the `_parse_clixml` it's required.
image

felixfontein (approved):
The non-Windows part of this PR looks good (i.e. it should not break anything ;) ).

plugins/connection/docker.py
```diff
@@ -308,11 +330,32 @@ def put_file(self, in_path, out_path):
                 count = ' count=0'
             else:
                 count = ''
-            args = self._build_exec_cmd([self._play_context.executable, "-c", "dd of=%s bs=%s%s" % (out_path, BUFSIZE, count)])
+            if getattr(self._shell, "_IS_WINDOWS", False):
+                cmd = '''& { $strm = [Console]::OpenStandardInput(1024)
+$data = [System.Array]::CreateInstance([byte],1024)
+$fstrm = [IO.File]::OpenWrite("%s")
```
briantist, line 336:
```suggestion
$fstrm = [IO.File]::OpenWrite('%s')
```
Possibly a nit, but on the off-chance `out_path` contains `$`...

tmeckel:
@felixfontein Why are the RHEL 8.2 tasks in the CI pipeline are failing. Is this relevant to my PR?

changelogs/fragments/96-windows-support.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - docker_container - added support for Windows containers (https://github.com/ansible-collections/community.docker/pull/96).
```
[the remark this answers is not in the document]
tmeckel: I only carried over @felixfontein proposal

plugins/connection/docker.py
tmeckel: @braintist eagle eye! 😮

changelogs/fragments/96-windows-support.yaml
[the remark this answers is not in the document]
felixfontein: I must have been somehow sleeping... Sorry :)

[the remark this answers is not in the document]
felixfontein:
```suggestion
  - docker connection plugin - added support for Windows containers (https://github.com/ansible-collections/community.docker/pull/96).
```

felixfontein:
@tmeckel the RHEL issues are unrelated to this PR, there seems to be something wrong with the setup of the VMs. I've pinged someone to investigate.

felixfontein:
The RHEL will be fixed by #97.

felixfontein:
Tests should now pass. Let's see :)

felixfontein:
recheck

felixfontein:
@briantist @jborean93 I plan to create a new release of community.docker on Monday or Tuesday next week (so it can get included in Ansible 3.1.0). If you don't manage to review the PR until then, I can also create another release once this is merged, so no pressure ;)

briantist:
> I plan to create a new release of community.docker on Monday or Tuesday next week (so it can get included in Ansible 3.1.0). If you don't manage to review the PR until then, I can also create another release once this is merged, so no pressure ;)

I'm not sure I'll be able to actually run this to try it out unfortunately, certainly not before your next release. I'll see if I get a chance to try it out at some point

plugins/connection/docker.py
```diff
@@ -268,6 +274,12 @@ def exec_command(self, cmd, in_data=None, sudoable=False):
         display.debug("done communicating")
 
         display.debug("done with docker.exec_command()")
+
+        # When running on Windows, stderr may contain CLIXML encoded output
+        if getattr(self._shell, "_IS_WINDOWS", False) and stderr.startswith(b"#< CLIXML"):
+            from ansible.plugins.shell.powershell import _parse_clixml
```
jborean93, line 280: I would be concerned with importing this function, it's currently a "private" function and the location has changed in the past before. You might be better off just copying the function here rather than trying to import this then find out a future Ansible version has changed it once again.

plugins/connection/docker.py
```diff
@@ -214,7 +216,11 @@ def exec_command(self, cmd, in_data=None, sudoable=False):
         """ Run a command on the docker host """
         super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
 
-        local_cmd = self._build_exec_cmd([self._play_context.executable, '-c', cmd])
+        if getattr(self._shell, "_IS_WINDOWS", False):
+            sudoable = False
+            local_cmd = self._build_exec_cmd(["cmd", "/c", cmd])
```
jborean93, line 221: I might be wrong but I believe using `cmd /c` here will mean any `raw` task with the connection plugin will be using `cmd` as the shell rather than PowerShell. You may want to change this based on whether the `ansible_shell_type` is either `cmd` or `powershell` to allow a user to change this as they desire.

plugins/connection/docker.py
```diff
@@ -308,11 +330,32 @@ def put_file(self, in_path, out_path):
...
+do {
+    $read = $strm.Read($data, 0, $data.Length)
+    if ($read -gt 0) {
+        $fstrm.Write($data, 0, $read)
+    }
+} while ($read -gt 0)
+$fstrm.Flush()
+$fstrm.Dispose()
+$strm.Dispose() }''' % (out_path)
+                args = self._build_exec_cmd(["cmd", "/c", self._shell._encode_script(cmd, as_list=False, strict_mode=False, preserve_rc=False)])
```
jborean93, line 346: This will start `cmd /c` to then call `powershell`. You may as well just start `powershell` direct to make it more efficient

```suggestion
                args = self._build_exec_cmd(self._shell._encode_script(cmd, as_list=True, strict_mode=False, preserve_rc=False))
```

plugins/connection/docker.py
```diff
@@ -308,11 +330,32 @@ def put_file(self, in_path, out_path):
                 count = ' count=0'
             else:
                 count = ''
-            args = self._build_exec_cmd([self._play_context.executable, "-c", "dd of=%s bs=%s%s" % (out_path, BUFSIZE, count)])
+            if getattr(self._shell, "_IS_WINDOWS", False):
+                cmd = '''& { $strm = [Console]::OpenStandardInput(1024)
```
jborean93, line 334: If you use `_encode_script` as per below you can skip the `& { ... }` entirely and just run the command as is.

plugins/connection/docker.py
```diff
@@ -308,11 +330,32 @@ def put_file(self, in_path, out_path):
...
+$data = [System.Array]::CreateInstance([byte],1024)
+$fstrm = [IO.File]::OpenWrite(%s)
```
jborean93, line 336: You still need to have more logic to handle here. Right now at a quote glance I could have a path like:

```
’+ (&[ScriptBlock]::Create([[Text.Encoding]::UTF8.GetString([Convert]::FromBase64String("my code to inject")) +’
```
_Note the use of unicode smart quotes_

PowerShell (un)helpfully treats a bunch of smart quotes as the same as single quotes so even if you ran `shlex_quote` on your path it won't escape what is needed. As soon as you can escape outside of a quote you can embed whatever command you want to run here.

There are 2 choices I can see you moving to:

* Use a the same (or similar regex) to the ansible.windows [quote filter for powershell](https://github.com/ansible-collections/ansible.windows/blob/main/plugins/filter/quote.py#L63-L70) to escape the smart quotes and give you a literal value
* Base64 encode the path and decode that in your script - base64 means you no longer have to deal with quotes or unicode characters

```python
cmd = '''
...
$outPath = [Text.Encoding]::UTF8.GetString([Convert]::FromBase64String('%s'))
...
''' % (to_text(base64.b64encode(to_bytes(out_path, encoding='utf-8'))))

Both options remove the need for `shlex_quote` but I still recommend you normalize the path as you have done today.

**Review by jborean93 — COMMENTED** [review]
Some nice edge case stuff in the review but what you have here is pretty good.

I will highly recommend that tests be added as soon as possible, there are a few private functions and edge cases with Windows that are hard to keep track off without tests actually testing those scenarios.

**Comment by felixfontein** [comment]
> I will highly recommend that tests be added as soon as possible, there are a few private functions and edge cases with Windows that are hard to keep track off without tests actually testing those scenarios.

There's one big problem with Windows container tests: someone has to create **and maintain** them. Since Windows containers cannot run on Linux, this requires the tests to run on Windows. Our CI (AZP) theoretically does support that, but we've never run tests on Windows for any of the docker plugins and modules yet. So this is far from simple to set up.

**Comment by jborean93** [comment]
Totally understand that problem, at a quick glance you could do what we did for testing SSH against Windows hosts. Still spin up an AWS EC2 instance for Windows Server 2019 and install the Containers feature and a running container. I'm not sure if it's possible to even expose the Docker port over the internet like that so that might be a no go. Ultimately it's up to you as the maintainers what you would like to have. I just know this is going to be painful to support going forwards without some kind of tests around it.

**Comment by jimbo8098** [comment]
If this were to be accepted, I have a CI pipeline for my organisation which would make use of these utilities for docker swarm creation and more. We use WinRM and Windows Server 2019 predominantly but resources to work with others since idempotent IasC is in it's infancy here.

I was actually on my way to find out the views of me PRing the `community.docker.docker_swarm` module to support Windows!

**Comment by felixfontein** [comment]
@jimbo8098 I'm happy to merge this (once all review comments are addressed) also without Windows tests. We (or in particularly I :-) ) don't mind Windows-specific PRs - as long as they don't break other platforms.

It would be great to have Windows tests in this repository as well; as long as we don't have them, Windows support in the modules is "use on your own risk, we try not to break it, but we won't notice in case we do". If your CI pipeline uses say the `main` branch of this collection, you would notice early if something breaks and you could create an issue (or even a PR to fix it), so that releases should usually work for Windows :)

**Comment by jimbo8098** [comment]
Just to add to my previous point, Our infrastructure is Windows Server 2019 hardware, not containers. I suspect the two should be the same in a CI environment if you are just building towards testing infra for this though some mechanisms of course might be a bit different. Will look to PR on `docker_swarm` for the moment but I'll be watching this.

**Comment by felixfontein** [comment]
@tmeckel do you have time to incorporate the latest review comments? It would be great if we could get this merged.

**Comment by tmeckel** [comment]
@felixfontein sorry for the delay. Found some time to work on the comments of @jborean93

**Comment by felixfontein** [comment]
@tmeckel ping

File: plugins/connection/docker.py (hunk 6 above)
Comment (tmeckel) [reply plugins/connection/docker.py:280]: @jborean93 copied `_parse_clixml` to `docker-py`

File: plugins/connection/docker.py (hunk 7 above)
Comment (tmeckel) [reply plugins/connection/docker.py:221]: @jborean93 yes this is true. Bit when I change the code as follows I receive a `file not found`. So the interpreter (Powershell) cannot be found. 

image

File: plugins/connection/docker.py (hunk 9 above)
Comment (tmeckel) [reply plugins/connection/docker.py:334]: @jborean93 changed the code as advised, but I receive an `invalid payload`

image


image

**Comment by tmeckel** [comment]
@felixfontein added comments. Awaiting answers from @jborean93

File: plugins/connection/docker.py (hunk 7 above)
Comment (jborean93) [reply plugins/connection/docker.py:221]: Do you know what the default shell is for a Windows container, i.e. what runs when you do `docker exec ...`. Is it cmd or is it powershell.exe, or maybe no shell at all?

Basically right now if someone was to do the following

```yaml
- raw: echo "hi"
```

Is that going to be running under cmd or powershell? With the current changes here that will be running as `cmd` which might not be ideal if someone has set `ansible_shell_type=powershell`. Maybe the desire is to have `raw` with this connection plugin independent to the shell plugin used, if that's the case then the PR as it is makes sense. If you wish for this to be dependent on the shell plugin then you will have to hardcode `cmd.exe /c` when using the cmd shell and `powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -Command` for the powershell shell.

This might be something that @felixfontein may want to comment on.

plugins/connection/docker.py
jborean93: I'm not sure what is happening here, the error you've shared is something that is thrown by the [bootstrap_wrapper.ps1](https://github.com/ansible/ansible/blob/devel/lib/ansible/executor/powershell/bootstrap_wrapper.ps1) which is used when executing a module. The code you have here is for copying a file to the docker container and shouldn't be calling the bootstrap_wrapper code.

plugins/connection/docker.py
felixfontein: I don't think I can say something here. I've never used Docker with Windows containers, so I have zero experience what these contain.

jborean93: I mostly mean more in general, do you expect that `raw` should match the `ansible_shell_type` (powershell or cmd) or should it just be the same thing on the connection plugin?

felixfontein: Matching `ansible_shell_type` would be nice I guess, though I'm not even sure whether Linux containers fully support that (I've never really looked at shell plugins). I also have no idea how complicated it would be to 'properly' implement this.

PR #103, opened 2021-03-12
[security] docker_swarm: mark join_token as no_log

felixfontein:
##### SUMMARY
The `join_token` parameter of `docker_swarm` is currently not marked as `no_log`, which causes its value to be written to syslog during module invocation. This PR changes that.

An unfortunate side-effect is that if `join_token` is specified, the value there will be censored from the return values (i.e. replaced by `VALUE_SPECIFIED_IN_NO_LOG_PARAMETER`). This can potentially break playbooks/roles which store the return value.

This **only** affects the situation when `join_token` is provided; I think that usually in that case, the return values are not used, so it should break relatively few things (if any).

CC @WojciechowskiPiotr @relrod

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
docker_swarm

### DIFF
--- a/plugins/modules/docker_swarm.py
+++ b/plugins/modules/docker_swarm.py
@@ -83,6 +83,8 @@ options:
     description:
       - Swarm token used to join a swarm cluster.
       - Used with I(state=join).
+      - If this value is specified, the corresponding value in the return values will be censored by Ansible.
+        This is a side-effect of this value not being logged.
     type: str
   remote_addrs:
     description:
@@ -230,12 +232,20 @@ swarm_facts:
           type: dict
           contains:
               Worker:
-                  description: Token to create a new *worker* node
+                  description:
+                    - Token to create a new *worker* node.
+                    - "B(Note:) if this value has been specified as I(join_token), the value here will not
+                       be the token, but C(VALUE_SPECIFIED_IN_NO_LOG_PARAMETER). If you pass I(join_token),
+                       make sure your playbook/role does not depend on this return value!"
                   returned: success
                   type: str
                   example: SWMTKN-1--xxxxx
               Manager:
-                  description: Token to create a new *manager* node
+                  description:
+                    - Token to create a new *manager* node.
+                    - "B(Note:) if this value has been specified as I(join_token), the value here will not
+                       be the token, but C(VALUE_SPECIFIED_IN_NO_LOG_PARAMETER). If you pass I(join_token),
+                       make sure your playbook/role does not depend on this return value!"
                   returned: success
                   type: str
                   example: SWMTKN-1--xxxxx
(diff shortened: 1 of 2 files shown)


plugins/modules/docker_swarm.py
```diff
@@ -230,12 +232,20 @@
...
+                    - "B(Note:) if this value has been specified as I(join_token), the value here will not
+                       be the token, but C(VALUE_SPECIFIED_IN_NO_LOG_PARAMETER). If you pass I(join_token),
+                       make sure your playbook/role does not depend on this return value!"
                   returned: success
                   type: str
                   example: SWMTKN-1--xxxxx
               Manager:
-                  description: Token to create a new *manager* node
+                  description:
+                    - Token to create a new *manager* node.
```
WojciechowskiPiotr, line 245:
```suggestion
                    - Token to join the cluster as a new *manager* node.
```

plugins/modules/docker_swarm.py
```diff
@@ -230,12 +232,20 @@
           type: dict
           contains:
               Worker:
-                  description: Token to create a new *worker* node
+                  description:
+                    - Token to create a new *worker* node.
```
WojciechowskiPiotr, line 236:
```suggestion
                    - Token to join the cluster as a new *worker* node.
```

WojciechowskiPiotr (commented):
What options we have to get the tokens now? Maybe we should also add the comment to get the tokens you have to use `docker_swarm_info` module? As far as I see it is not censored there in output.

felixfontein:
> What options we have to get the tokens now? Maybe we should also add the comment to get the tokens you have to use `docker_swarm_info` module? As far as I see it is not censored there in output.

Exactly the same as before - except if you call the module with `join_token`, which you only do if you already have a token. Then you won't get the same token back in the result.

WojciechowskiPiotr:
> > What options we have to get the tokens now? Maybe we should also add the comment to get the tokens you have to use `docker_swarm_info` module? As far as I see it is not censored there in output.
> 
> Exactly the same as before - except if you call the module with `join_token`, which you only do if you already have a token. Then you won't get the same token back in the result.

So tokens are hidden in the `results` when you create the cluster?

felixfontein:
> > > What options we have to get the tokens now? Maybe we should also add the comment to get the tokens you have to use `docker_swarm_info` module? As far as I see it is not censored there in output.
> > 
> > 
> > Exactly the same as before - except if you call the module with `join_token`, which you only do if you already have a token. Then you won't get the same token back in the result.
> 
> So tokens are hidden in the `results` when you create the cluster?

When you create a cluster, you do not specify `join_token` (since you are not joining), right? In that case the tokens will be returned.

felixfontein:
@WojciechowskiPiotr thanks for reviewing and merging! I'll work on backports later today.

PR #114, opened 2021-03-26
Add use_ssh_client flag

### LINKED ISSUE #108: Feature Request: support "use_ssh_client" parameter for Docker SSH connections
In attempting to use the community.docker collection with a Docker host I access via SSH, I have found that the current reliance on Paramiko is flaky, and does not support a lot of edge cases for SSH authentication. On doing a bit of research, I have found that the Docker SDK for Python recently added a new parameter `use_ssh_client` to allow the use of the local ssh client instead of the Paramiko module (see documentation [here](https://docker-py.readthedocs.io/en/stable/client.html#docker.client.DockerClient)).

I think that this is called here:
https://github.com/ansible-collections/community.docker/blob/f2d16da6435cb3bc493bbe2efdf50371e5ea503a/plugins/module_utils/common.py#L261-L264

I made a quick and dirty change to these lines to try it out (by adding `use_ssh_client=True` to the dictionary), and it seems to work on my local machine! I'd like to submit a pull request - however, I've no idea what other changes I'd need to make in order to make this a configurable parameter instead of hard-coded, or what testing/regression testing I'd need to do to validate it didn't screw some other setup up!

Looking for advice/guidance from the group about how best to proceed - or happy for someone else to take this and run with it 🙂.

felixfontein:
##### SUMMARY
Resolves #108.

This has no tests since we currently don't have a good way to test this. I will think more about this tomorrow.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
most modules and plugins

### DIFF
--- a/changelogs/fragments/114-use_ssh_client.yml
+++ b/changelogs/fragments/114-use_ssh_client.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- "Add the ``use_ssh_client`` option to most docker modules and plugins (https://github.com/ansible-collections/community.docker/issues/108, https://github.com/ansible-collections/community.docker/pull/114)."
--- a/plugins/doc_fragments/docker.py
+++ b/plugins/doc_fragments/docker.py
@@ -80,6 +80,13 @@ options:
               instead. If the environment variable is not set, the default value will be used.
         type: bool
         default: no
+    use_ssh_client:
+        description:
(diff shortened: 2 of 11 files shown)


felixfontein:
ready_for_review

MisterWS:
As mentioned in #108, LGTM from my limited testing.

felixfontein:
@MisterWS thanks a lot for testing this!

PR #129, opened 2021-04-27
Document the issues with tls_hostname

larsks:
Add notes to the module documentation making it explicit that `tls_hostname` must be set when connecting to a remote Docker service.

See #128 for additional details.

### DIFF
--- a/plugins/doc_fragments/docker.py
+++ b/plugins/doc_fragments/docker.py
@@ -15,8 +15,9 @@ options:
     docker_host:
         description:
             - The URL or Unix socket path used to connect to the Docker API. To connect to a remote host, provide the
-              TCP connection string. For example, C(tcp://192.0.2.23:2376). If TLS is used to encrypt the connection,
-              the module will automatically replace C(tcp) in the connection URL with C(https).
+              TCP connection string. For example, C(tcp://192.0.2.23:2376).
+            - If TLS is used to encrypt the connection, the module will automatically replace C(tcp) in the connection
+              URL with C(https). See the documentation on I(tls_hostname) for additional information about using TLS.
             - If the value is not specified in the task, the value of environment variable C(DOCKER_HOST) will be used
               instead. If the environment variable is not set, the default value will be used.
         type: str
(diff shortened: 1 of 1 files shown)


plugins/doc_fragments/docker.py
```diff
@@ -27,6 +28,9 @@ class ModuleDocFragment(object):
             - When verifying the authenticity of the Docker Host server, provide the expected name of the server.
             - If the value is not specified in the task, the value of environment variable C(DOCKER_TLS_HOSTNAME) will
               be used instead. If the environment variable is not set, the default value will be used.
+            - Due to an implementation error in this module, if you are connecting to a host other than C(localhost)
+              using TLS, you B(must) set this parameter (or C(DOCKER_TLS_HOSTNAME)) to the remote hostname.  This issue
+              will be corrected in a future version of this module.
```
felixfontein, line 33: It is **not** an error, it is an unfortunate design choice. Also there is no guarantee that a docker daemon running on `localhost` has a certificate which includes `localhost`. And finally, this is not necessary a module, but the same doc fragment is also used in plugins.

Besides that, all necessary information is already given in the first part of the description.
```suggestion
            - If you obtain errors such as C(hostname 'localhost' doesn't match
'docker.local'), you probably have to set this option to the correct value.
```

plugins/doc_fragments/docker.py
```diff
@@ -15,8 +15,9 @@ class ModuleDocFragment(object):
     docker_host:
         description:
             - The URL or Unix socket path used to connect to the Docker API. To connect to a remote host, provide the
-              TCP connection string. For example, C(tcp://192.0.2.23:2376). If TLS is used to encrypt the connection,
-              the module will automatically replace C(tcp) in the connection URL with C(https).
+              TCP connection string. For example, C(tcp://192.0.2.23:2376).
+            - If TLS is used to encrypt the connection, the module will automatically replace C(tcp) in the connection
+              URL with C(https). See the documentation on I(tls_hostname) for additional information about using TLS.
```
felixfontein, line 20:
```suggestion
              URL with C(https). When using certificate validation (see the I(validate_certs) option), you probably
              must also provide I(tls_hostname).
```

PR #138, opened 2021-05-10
Allow memory_swap=-1 (unlimited)

shane-axiom:
`--memory-swap=-1` is a valid docker run value allowing
unlimited swap usage. `docker_container` should allow
this value as it sometimes must explicitly be set to prevent
docker from attempting to limit swap usage to double the
`--memory` value by default. In rootless docker environments,
attempting to set `memory-swap` usage limits will result in an error
if the user doesn't have the ability to do so. Allowing -1 here
will prevent these errors.

https://docs.docker.com/config/containers/resource_constraints/#--memory-swap-details

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`docker_container`


changelogs/fragments/138-docker_container-allow-memory-swap-unlimited.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
felixfontein, line 1: This is more a feature, not a bugfix:
```suggestion
minor_changes:
```

changelogs/fragments/138-docker_container-allow-memory-swap-unlimited.yml
```diff
@@ -0,0 +1,2 @@
...
+- "docker_container - allow ``memory_swap: -1``, as it is a valid ``docker run`` value allowing unlimited swap usage. This is useful when the user cannot set memory swap limits due to cgroup limitations etc, as by default Docker will try to set swap usage to 2x ``memory`` (https://github.com/ansible-collections/community.docker/pull/138)"
```
felixfontein, line 2:
```suggestion
- "docker_container - allow ``memory_swap: -1`` to set memory swap limit to unlimited. This is useful when the user cannot set memory swap limits due to cgroup limitations or other reasons, as by default Docker will try to set swap usage to two times the value of ``memory`` (https://github.com/ansible-collections/community.docker/pull/138)."
```

plugins/modules/docker_container.py
```diff
@@ -459,8 +459,9 @@
   memory_swap:
     description:
       - "Total memory limit (memory + swap) in format C(<number>[<unit>]).
-        Number is a positive integer. Unit can be C(B) (byte), C(K) (kibibyte, 1024B),
-        C(M) (mebibyte), C(G) (gibibyte), C(T) (tebibyte), or C(P) (pebibyte)."
+        Number is a positive integer or -1 (unlimited). Unit can be C(B) (byte),
```
felixfontein, line 462: How about using the value `unlimited` instead of `-1` for this?

plugins/modules/docker_container.py
```diff
@@ -1412,7 +1413,9 @@ def __init__(self, client):
             self.groups = [to_text(g, errors='surrogate_or_strict') for g in self.groups]
 
         for param_name in REQUIRES_CONVERSION_TO_BYTES:
-            if client.module.params.get(param_name):
+            #don't convert memory_swap=-1 as it's a valid value (allows unlimited swap usage)
+            if client.module.params.get(param_name) and not (param_name == 'memory_swap'
+                and client.module.params.get(param_name) == '-1'):
```
felixfontein, line 1418: I would rather write this as:
```.py
             if client.module.params.get(param_name):
                 if param_name == 'memory_swap' and client.module.params[param_name] == 'unlimited':
                     setattr(self, param_name, -1)
                 else:
                     try:
                         ...
```

felixfontein:
BTW: there's a release planned for tomorrow, so if we can get this PR in a mergeable shape in time, it can get included in the 1.6.0 release.

plugins/modules/docker_container.py
shane-axiom: @felixfontein Any thoughts on allowing both `-1` and `unlimited`? `unlimited` is definitely clearer, but `-1` is used by docker itself.

felixfontein (approved):

tests/integration/targets/docker_container/tasks/tests/options.yml
```diff
@@ -2534,6 +2560,10 @@ avoid such warnings, please quote the value.' in log_options_2.warnings"
     # on a proper VM), memory_swap cannot be set and will be -1 afterwards.
     - memory_swap_2 is not changed or memory_swap_2.container.HostConfig.MemorySwap == -1
     - memory_swap_3 is changed
+    # Unlimited memory_swap (using 'unlimited') should be allowed
+    - memory_swap_4 is changed and memory_swap_4.container.HostConfig.MemorySwap == -1
```
felixfontein, line 2564:
```suggestion
    # (If the value was already -1 because of the above reasons, it won't change)
    - (memory_swap_4 is changed or memory_swap_3.container.HostConfig.MemorySwap == -1) and memory_swap_4.container.HostConfig.MemorySwap == -1
```
This is not surprising (from the above comments). The tests passed on the VMs (the RHEL CI nodes are VMs, the others are docker containers), so they work fine in principle :)

plugins/modules/docker_container.py
```diff
@@ -459,8 +459,9 @@
   memory_swap:
     description:
       - "Total memory limit (memory + swap) in format C(<number>[<unit>]).
-        Number is a positive integer. Unit can be C(B) (byte), C(K) (kibibyte, 1024B),
-        C(M) (mebibyte), C(G) (gibibyte), C(T) (tebibyte), or C(P) (pebibyte)."
+        Number is a positive integer or 'unlimited'/-1 for unlimited swap usage.
```
felixfontein, line 462:
```suggestion
        Number is a positive integer, or C(unlimited) or C(-1) for unlimited swap usage.
```

plugins/modules/docker_container.py
```diff
@@ -459,8 +459,9 @@
   memory_swap:
     description:
       - "Total memory limit (memory + swap) in format C(<number>[<unit>]).
```
felixfontein, line 461:
```suggestion
      - "Total memory limit (memory + swap) in format C(<number>[<unit>]), or
        the special values C(unlimited) or C(-1) for unlimited swap usage.
```

plugins/modules/docker_container.py
```diff
@@ -459,8 +459,9 @@
...
-        Number is a positive integer. Unit can be C(B) (byte), C(K) (kibibyte, 1024B),
-        C(M) (mebibyte), C(G) (gibibyte), C(T) (tebibyte), or C(P) (pebibyte)."
+        Number is a positive integer or 'unlimited'/-1 for unlimited swap usage.
+        Unit can be C(B) (byte), C(K) (kibibyte, 1024B), C(M) (mebibyte), C(G) (gibibyte),
+        C(T) (tebibyte), or C(P) (pebibyte)."
```
felixfontein, line 464:
```suggestion
        Number is a positive integer. Unit can be C(B) (byte), C(K) (kibibyte, 1024B),
        C(M) (mebibyte), C(G) (gibibyte), C(T) (tebibyte), or C(P) (pebibyte)."
```

felixfontein:
@shane-axiom thanks for implementing this! :)

PR #143, opened 2021-05-16
Only emit tls_hostname deprecation warning if TLS is actually used

felixfontein:
##### SUMMARY
Right now it is also emitted if TLS is not used to talk to Docker daemon.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/common.py


ssbarnea:
@felixfontein I observed these really annoying warnings with molecule-docker driver and I failed to find any combination of arguments to give to docker modules that would avoid creating these warnings.

That is weird because I usually use a remove docker host using SSH, like `DOCKER_HOST=ssh://ssbarnea@example.com` which is secure by definition and still not TLS related.

Note: disabling all Ansible warnings is no option, it would be a very irresponsible move.

ssbarnea (approved):

ssbarnea:
I confirm that this fix address the issue for for those using ssh:// too.

felixfontein:
@ssbarnea thanks for testing/reviewing!

PR #149, opened 2021-06-04
docker_image: allow to tag images by name, improve docs

### LINKED ISSUE #147: Support re-tagging docker image
##### SUMMARY

I need to retag an image, or to issue this command, basically: `docker image tag $source $destination`. No way to do it with `community.docker.docker_image` AFAICS.

The purpose is to roll back to an older image after a failed deployment.

##### ISSUE TYPE
- The `source` argument doesn't allow specifying a different image as source.
- A new `local_source` may help.

##### COMPONENT NAME

docker_image

##### ADDITIONAL INFORMATION

```yaml
- name: re-tag old image
  community.docker.docker_image:
    source: local
    local_source: "{{ old_image }}"
    force_tag: yes
    name: deployed/image/name:tag
```

felixfontein:
##### SUMMARY
Fixes #147.

Also adds explicit tests which tag a docker image by ID.

##### ISSUE TYPE
- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME
docker_image

### DIFF
--- a/plugins/modules/docker_image.py
+++ b/plugins/modules/docker_image.py
@@ -624,7 +624,7 @@ class ImageManager(DockerBaseClass):
 
         if not image or self.force_tag:
             image_name = name
-            if not is_image_name_id(name) and tag and re.search(tag, name):
+            if not is_image_name_id(name) and tag and not re.search(tag, name):
                 image_name = "%s:%s" % (name, tag)
             self.log("tagging %s to %s:%s" % (image_name, repo, repo_tag))
             self.results['changed'] = True
--- a/plugins/modules/docker_image.py
+++ b/plugins/modules/docker_image.py
@@ -590,8 +590,8 @@ class ImageManager(DockerBaseClass):
                             changed = True
                     self.results['changed'] = changed
                 except Exception as exc:
-                    if re.search('unauthorized', str(exc)):
-                        if re.search('authentication required', str(exc)):
+                    if 'unauthorized' in str(exc):
+                        if 'authentication required' in str(exc):
                             self.fail("Error pushing image %s/%s:%s - %s. Try logging into %s first." %
                                       (registry, repo_name, tag, to_native(exc), registry))
                         else:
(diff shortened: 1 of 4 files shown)


plugins/modules/docker_image.py
```diff
@@ -179,8 +180,10 @@
     default: no
   repository:
     description:
-      - Full path to a repository. Use with state C(present) to tag the image into the repository. Expects
-        format I(repository:tag). If no tag is provided, will use the value of the C(tag) parameter or I(latest).
+      - Use with state C(present) to tag the image.
+      - Expects format C(repository:tag). If no tag is provided, will use the value of the I(tag) parameter or C(latest).
+      - If I(push=true), I(repository) must either include a registry, or will be assumed to belong to the default
```
aminvakil, line 185: I may misunderstood what you have meant by including a registry though.
```suggestion
      - If I(push=true), I(repository) must either be a registry, or will be assumed to belong to the default
```

felixfontein: A registry is "your.registry.com:5000", a name including a registry is "your.registry.com:5000/repo/image". I mean the latter here.

aminvakil: Ah, sure, thanks for clarification.

felixfontein:
Ok, this is no longer a docs PR, but a feature + docs PR. It allows to tag images by ID, which didn't work before because it tried to add `:tag` to the image ID... (Since we didn't document before that it should work, it's a feature and not a bugfix.)

felixfontein:
ready_for_review

plugins/modules/docker_image.py
```diff
@@ -620,16 +623,16 @@ def tag_image(self, name, tag, repository, push=False):
...
+            self.log("tagging %s to %s:%s" % (image_name, repo, repo_tag))
             self.results['changed'] = True
-            self.results['actions'].append("Tagged image %s:%s to %s:%s" % (name, tag, repo, repo_tag))
+            self.results['actions'].append("Tagged image %s to %s:%s" % (image_name, repo, repo_tag))
             if not self.check_mode:
                 try:
                     # Finding the image does not always work, especially running a localhost registry. In those
                     # cases, if we don't set force=True, it errors.
-                    image_name = name
-                    if tag and not re.search(tag, name):
```
briantist, line 631: couldn't leave this comment on a line not included in the PR it seems, but unless I missed any, you've removed all the regex uses, so should be able to remove `import re` I think?

briantist (approved):
looks good to me, only small thing is that it looks like you can get rid of the `re` import now.

plugins/modules/docker_image.py
felixfontein: Thanks, fixed in dc8550dcb511a3faeebecda40b33a036e5d3f08b.

felixfontein:
@aminvakil @briantist thanks a lot for reviewing!

PR #150, opened 2021-06-06
Add basic docker_compose tests

felixfontein:
##### SUMMARY
So we have more than nothing.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
docker_compose


aminvakil:
AFAIK `docker-compose` is not available on official repositories or it is too old.
What do you think about installing them from github directly?
Something like this:
```
---
- name: Get docker-compose checksum
  get_url:
    url: https://github.com/docker/compose/releases/download/{{ docker_compose_version }}/docker-compose-{{ ansible_system }}-{{ ansible_architecture }}.sha256
    dest: /tmp/docker_compose_checksum
    force: true

- name: Get sha256sum from docker-compose checksum file
  slurp:
    src: /tmp/docker_compose_checksum
  register: docker_compose_checksum_encoded

- name: Decode docker_compose_checksum_encoded
  set_fact:
    docker_compose_checksum: '{{ docker_compose_checksum_encoded["content"] | b64decode }}'

- name: Install docker-compose
  get_url:
    url: https://github.com/docker/compose/releases/download/{{ docker_compose_version }}/docker-compose-{{ ansible_system }}-{{ ansible_architecture }}
    dest: /usr/local/bin/docker-compose
    mode: '0755'
    owner: root
    group: root
    checksum: 'sha256:{{ docker_compose_checksum.split(" ")[0] }}'
    force: true
```
https://github.com/aminvakil/ansible-role-docker-installation/blob/master/tasks/docker-compose.yml

felixfontein:
> AFAIK `docker-compose` is not available on official repositories or it is too old.

Well, that's exactly the point of using system packages :) On the systems where it is installed with `pip`, we already get the latest version.

felixfontein:
> > AFAIK `docker-compose` is not available on official repositories or it is too old.
> 
> Well, that's exactly the point of using system packages :) On the systems where it is installed with `pip`, we already get the latest version.

aminvakil:
> > AFAIK `docker-compose` is not available on official repositories or **it is too old**.
> 
> Well, that's exactly the point of using system packages :) On the systems where it is installed with `pip`, we already get the latest version.

I have not suggested using **latest version** from github, it's just that I've seen `docker-compose` package in official repositories act buggy :) Therefore I've switched to using github.

Although if they work here and tests pass, why not?

felixfontein:
I'm not completely happy with the current setup, and there are too many places where I just disabled tests in the end, but I'm going to merge this anyway so we have something we can start working with and something we can improve :) The main problem is that docker-compose has very specific Python dependencies, so installing by `pip` is quite demanding, and installing system packages is also problematic in case Pypi packages have already installed for some of the requirements it needs.

But at least some very basic tests for docker_compose are running now, and can be improved. :)

PR #157, opened 2021-06-21
Pass Docker daemon connection params from inventory to connection plugin

### LINKED ISSUE #155: Inventory plugin should propagate docker_host variable
##### SUMMARY
I use inventory plugin with remote docker_host. However it is not propagated to the connection plugin and the latter uses default local connection.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.docker.docker_containers

##### ANSIBLE VERSION
```paste below
ansible 2.10.5
  config file = /home/default/dev-container/ansible.cfg
  configured module search path = ['/home/default/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.9.4 (default, Apr  4 2021, 19:38:44) [GCC 10.2.1 20210401]
```

##### CONFIGURATION
```paste below
DEFAULT_HOST_LIST(/home/default/dev-container/ansible.cfg) = ['/home/default/dev-container/inventory']
```

##### OS / ENVIRONMENT
Ubuntu 21.04

##### STEPS TO REPRODUCE
Inventory:
```yaml
plugin: community.docker.docker_containers
docker_host: tcp://my-host:2376
```

##### EXPECTED RESULTS
I can use those hosts in my playbook, they should connect to `my-host:2376` 

##### ACTUAL RESULTS
Playbook fails with:
```
TASK [Gathering Facts] **********************************************************************************************************************************************************************************************************************************
fatal: [dev-vm]: UNREACHABLE! => {"changed": false, "msg": "Could not find container \"dev-vm\" (404 Client Error for http+docker://localhost/v1.40/containers/dev-vm/exec: Not Found (\"No such container: dev-vm\"))", "unreachable": true}
```

However ansible-inventory successfully finds containers:

```
$ ansible-inventory --graph --vars
@all:
  |--@ungrouped:
  |  |--dev-vm
  |  |  |--{ansible_connection = community.docker.docker_api}
  |  |  |--{ansible_host = dev-vm}
  |  |  |--{docker_name = dev-vm}
  |  |  |--{docker_short_id = e1a86cd6a5599}
```

##### Workaround

To work around I added variable explicitly to inventory:
```yaml
plugin: community.docker.docker_containers
docker_host: tcp://my-host:2376
compose:
  ansible_docker_docker_host: "'tcp://my-host:2376'"
```
But this looks like a hack and unnecessary duplication.

felixfontein:
##### SUMMARY
Fixes #155.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
docker_containers inventory plugin

### DIFF
--- a/plugins/inventory/docker_containers.py
+++ b/plugins/inventory/docker_containers.py
@@ -44,12 +44,22 @@ options:
               R(docker connection plugin,ansible_collections.community.docker.docker_connection),
               and C(docker-api) selects the
               R(docker_api connection plugin,ansible_collections.community.docker.docker_api_connection).
+            - When C(docker-api) is used, all Docker daemon configuration values are passed from the inventory plugin
+              to the connection plugin. This can be controlled with I(configure_docker_daemon).
         type: str
         default: docker-api
         choices:
             - ssh
             - docker-cli
             - docker-api
+
+    configure_docker_daemon:
+        description:
+            - Whether to pass all Docker daemon configuration from the inventory plugin to the connection plugin.
+            - Only used when I(connection_type=docker-api).
+        type: bool
+        default: true
+        version_added: 1.8.0
 
     verbose_output:
         description:
@@ -138,6 +148,7 @@ from ansible.plugins.inventory import BaseInventoryPlugin, Constructable
 
 from ansible_collections.community.docker.plugins.module_utils.common import (
     RequestException,
+    DOCKER_COMMON_ARGS_VARS,
 )
 from ansible_collections.community.docker.plugins.plugin_utils.common import (
     AnsibleDockerClient,
@@ -179,6 +190,13 @@ class InventoryModule(BaseInventoryPlugin, Constructable):
         if add_legacy_groups:
             self.inventory.add_group('running')
             self.inventory.add_group('stopped')
+
+        extra_facts = {}
+        if self.get_option('configure_docker_daemon'):
+            for option_name, var_name in DOCKER_COMMON_ARGS_VARS.items():
+                value = self.get_option(option_name)
+                if value is not None:
+                    extra_facts[var_name] = value
 
         for container in containers:
             id = container.get('Id')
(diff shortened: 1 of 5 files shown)


felixfontein:
According to @ytimenkov in https://github.com/ansible-collections/community.docker/issues/155#issuecomment-864976456, the PR solves #155.

plugins/inventory/docker_containers.py
```diff
@@ -44,13 +44,23 @@
...
             - ssh
             - docker-cli
             - docker-api
 
+    configure_docker_daemon:
+        description:
+            - Whether to pass all Docker daemon configuration from the inventory plugin to the connection plugin.
+            - Only used when I(connection_type=docker-api).
+        type: bool
+        default: true
```
briantist, line 61: I definitely agree the default should be `true` but it is a change in behavior isn't it? Would that consider it breaking?
Or is the current situation such that it couldn't ever be used effectively anyway (in which case there's no need to wait for major version)?

briantist (approved):
This looks pretty straightforward to me. Just one inline question.

plugins/inventory/docker_containers.py
felixfontein: I've thought a bit about it, and I'm still a bit torn; the original issue is a bug report, and I tend to also see the old behavior as a bug. I decided to also add this option to make it possible to revert to the old behavior, so in case someone really needs the old behavior they don't have to stick to an older collection version.

briantist: fwiw I agree with you; this is the logical/desired behavior, the previous behavior is a bug. It's always hard to tell when people have relied on buggy behavior. I'm not knowledgeable enough about this plugin to know how likely it is that the new default would have drastically (negative) different behavior; as an end user I would probably prefer something like this released sooner.

felixfontein:
@ytimenkov thanks for testing!
@briantist thanks for reviewing!

PR #160, opened 2021-06-21
docker_host_info - Allow filters which are passed as lists

### LINKED ISSUE #18: docker_host_info: Filter containers by multiple labels
##### SUMMARY

Cannot filter for multiple labels.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

docker_host_info

##### ADDITIONAL INFORMATION

This cannot work because we can't define the same key twice in a dict:

```yaml
- name: find running docker containers
  register: docker_info
  docker_host_info:
    containers: yes
    containers_filters:
      label: com.docker.compose.project=myproject # THIS GETS IGNORED
      label: com.docker.compose.service=backup
```

Possible solutions:

1.  Allow a list in `containers_filters`:

    ```yaml
    - name: find running docker containers
      register: docker_info
      docker_host_info:
        containers: yes
        containers_filters:
          - label=com.docker.compose.project=myproject
          - label=com.docker.compose.service=backup
    ```

1.  Allow a dict in `containers_filters.label` filter:

    ```yaml
    - name: find running docker containers
      register: docker_info
      docker_host_info:
        containers: yes
        containers_filters:
          label:
            com.docker.compose.project: myproject
            com.docker.compose.service: backup
    ```

I have not tested, but I guess the same applies to `images_filters`, `networks_filters` and `volumes_filters`.

Ajpantuso:
##### SUMMARY
Allows filter values which accept lists to be passed as lists to the docker API.
Fixes #18 
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME
plugins/modules/docker_host_info.py

##### ADDITIONAL INFORMATION
- Required an update to the common function `clean_dict_booleans_for_docker_api`, but previous behavior is defaulted so other modules are unaffected.
- integration tests for filters were added as well.

### DIFF
--- a/changelogs/fragments/160-docker_host_info-label-fitler-lists.yml
+++ b/changelogs/fragments/160-docker_host_info-label-fitler-lists.yml
@@ -1,3 +1,4 @@
 ---
 minor_changes:
-  - docker_host_info - allow values for keys in ``containers_filters`` to be passed as YAML lists (https://github.com/ansible-collections/community.docker/pull/160).   
+  - docker_host_info - allow values for keys in ``containers_filters``, ``images_filters``, ``networks_filters``, and
+    ``volumes_filters`` to be passed as YAML lists (https://github.com/ansible-collections/community.docker/pull/160).
--- a/changelogs/fragments/160-docker_host_info-label-fitler-lists.yml
+++ b/changelogs/fragments/160-docker_host_info-label-fitler-lists.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - docker_host_info - allow values for keys in ``containers_filters`` to be passed as YAML lists (https://github.com/ansible-collections/community.docker/pull/160).   
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -15,6 +15,7 @@ from distutils.version import LooseVersion
 
 
 from ansible.module_utils.basic import AnsibleModule, env_fallback, missing_required_lib
+from ansible.module_utils.common.collections import is_sequence
 from ansible.module_utils.common._collections_compat import Mapping, Sequence
 from ansible.module_utils.six import string_types
 from ansible.module_utils.six.moves.urllib.parse import urlparse
@@ -921,23 +922,27 @@ class DifferenceTracker(object):
         return result
 
 
-def clean_dict_booleans_for_docker_api(data):
+def clean_dict_booleans_for_docker_api(data, allow_sequences=False):
     '''
     Go doesn't like Python booleans 'True' or 'False', while Ansible is just
     fine with them in YAML. As such, they need to be converted in cases where
     we pass dictionaries to the Docker API (e.g. docker_network's
     driver_options and docker_prune's filters).
     '''
+    def sanitize_value(value):
+        if allow_sequences and is_sequence(value):
+            return type(value)(sanitize_value(v) for v in value)
+        elif value is True:
+            return 'true'
+        elif value is False:
+            return 'false'
+        else:
+            return str(value)
+
     result = dict()
     if data is not None:
         for k, v in data.items():
-            if v is True:
-                v = 'true'
-            elif v is False:
-                v = 'false'
-            else:
-                v = str(v)
-            result[str(k)] = v
+            result[str(k)] = sanitize_value(v)
     return result
 
 
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -927,12 +927,12 @@ def clean_dict_booleans_for_docker_api(data, allow_sequences=False):
     Go doesn't like Python booleans 'True' or 'False', while Ansible is just
     fine with them in YAML. As such, they need to be converted in cases where
     we pass dictionaries to the Docker API (e.g. docker_network's
-    driver_options and docker_prune's filters).
+    driver_options and docker_prune's filters). When `allow_sequences=True`
+    YAML sequences (lists, tuples) are converted to [str] instead of str([...])
+    which is the expected format of filters which accept lists such as labels.
     '''
-    def sanitize_value(value):
-        if allow_sequences and is_sequence(value):
-            return type(value)(sanitize_value(v) for v in value)
-        elif value is True:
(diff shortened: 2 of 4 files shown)


Ajpantuso:
SUSE should clear on recheck.

plugins/module_utils/common.py
```diff
@@ -921,23 +922,27 @@ def get_legacy_docker_diffs(self):
...
+def clean_dict_booleans_for_docker_api(data, allow_sequences=False):
     '''
     Go doesn't like Python booleans 'True' or 'False', while Ansible is just
     fine with them in YAML. As such, they need to be converted in cases where
     we pass dictionaries to the Docker API (e.g. docker_network's
     driver_options and docker_prune's filters).
     '''
+    def sanitize_value(value):
+        if allow_sequences and is_sequence(value):
+            return type(value)(sanitize_value(v) for v in value)
```
felixfontein, line 934: The API won't accept anything else than lists anyway, so it doesn't make to preserve the exact sequence type:
```suggestion
            return [sanitize_value(v) for v in value]
```

felixfontein: Another thing is that this code now processes lists recursively. So `[['a']]` will be kept, instead of being converted to `["['a']"]`. (The latter is not what the user wants, but what the API wants :) )

plugins/module_utils/common.py
```diff
@@ -921,23 +922,27 @@ def get_legacy_docker_diffs(self):
         return result
 
 
-def clean_dict_booleans_for_docker_api(data):
+def clean_dict_booleans_for_docker_api(data, allow_sequences=False):
     '''
     Go doesn't like Python booleans 'True' or 'False', while Ansible is just
     fine with them in YAML. As such, they need to be converted in cases where
     we pass dictionaries to the Docker API (e.g. docker_network's
     driver_options and docker_prune's filters).
```
felixfontein, line 930: Can you add something about lists when `allow_sequences=True` here?

plugins/modules/docker_host_info.py
```diff
@@ -126,6 +126,15 @@
     disk_usage: yes
   register: result
 
+- name: Get info on docker host and list containers matching the filter
```
felixfontein, line 129: You should also update the documentation for `containers_filters` and the other `_filters` options.

changelogs/fragments/160-docker_host_info-label-fitler-lists.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - docker_host_info - allow values for keys in ``containers_filters`` to be passed as YAML lists (https://github.com/ansible-collections/community.docker/pull/160).
```
felixfontein, line 3: Your change also affects the other `_filters` options.

felixfontein (commented):
Thanks for your contribution!

plugins/module_utils/common.py
Ajpantuso: Yeah, now that I look at the [API](https://github.com/docker/docker-py/blob/96c12726fdb9b24e985b4de74d5c82145ccd8185/docker/utils/utils.py#L366-L377)  they just convert every list item to strings so no recursion makes sense. I'll update this and the docs later on today.

plugins/modules/docker_host_info.py
```diff
@@ -33,6 +33,8 @@
     description:
       - A dictionary of filter values used for selecting containers to list.
       - "For example, C(until: 24h)."
+      - C(label) is a special case of filter which can be a string C(<key>) matching when a label is present, a string
+        C(<key>=<value>) matching when a label has a particular value, or a list of strings C(<key>)/C(<key>=<value).
```
felixfontein, line 37: Judging from some tests in docker-py, this might also work for exit codes (`exited: [0, 1]`; see https://github.com/docker/docker-py/blob/master/tests/unit/utils_test.py#L457). I wouldn't mention that in the docs without having tested it first, though...

Though when looking at the Docker daemon code, it looks like that's supported: https://github.com/moby/moby/blob/master/daemon/list.go#L247-L259 builds a list of accepted exit codes from all values passed for the `exited` filter, and https://github.com/moby/moby/blob/master/daemon/list.go#L463-L475 then checks containers for exit codes in that list.

Ajpantuso: I haven't studied these modules thoroughly, but I don't think we ever set `all=True` when calling `containers()` meaning we would only see running containers and `exited` would have no effect as a filter. Unless I'm way off.
```python
        try:
            if docker_object == 'containers':
                items = self.client.containers(**filter_arg)
```

Ajpantuso: To your point however when `all=True` a list of integers for `exited` works:
```yaml
- community.docker.docker_host_info:
         containers: true
         containers_filters:
           exited:
             - 137
             - 0
       register: result

     - debug:
         msg: "{{ result.containers }}"
```
partial result
```json
 {
            "Command": "/bin/bash -c 'sleep 1000000000000'",
            "Created": 1624312686,
            "Id": "a5f2a5adcf5a657221cbfb9d1e3612bdb6dc179d85938da46ed0b8c95004615a",
            "Image": "ubuntu",
            "Names": [
                "/crazy_mestorf"
            ],
            "Ports": [],
            "Status": "Exited (137) 22 hours ago"
        },
        {
            "Command": "/bin/bash",
            "Created": 1624311956,
            "Id": "cfdeb1c381ccd8dddf9dbd4eb11a633d6cda55b283d006dc5f9ce5d7b9fc879c",
            "Image": "ubuntu",
            "Names": [
                "/sweet_dewdney"
            ],
            "Ports": [],
            "Status": "Exited (0) 22 hours ago"
        }
```

felixfontein: Right, it needs `all=True`, otherwise there's no exit code ;-) I guess we can add support for `all=True` in a follow-up PR, but right now let's not document `exited`.

felixfontein (approved):

felixfontein:
@Ajpantuso thanks a lot for implementing this!

PR #162, opened 2021-06-25
docker_container - adding publish_all_ports option

### LINKED ISSUE #8: docker_container cannot use --publish and --publish-all at the same time
##### SUMMARY

When `--publish` and `--publish-all` are used at the same time, other ports except the specified port can be randomly allocated.
```
docker run -d -P -p 8080:8080 Nginx
```
Here, port 80 of Nginx will be randomly allocated.

```
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                                           NAMES
f7253236b603        nginx               "/docker-entrypoint.…"   5 minutes ago       Up 5 minutes        0.0.0.0:8080->8080/tcp, 0.0.0.0:32793->80/tcp   xenodochial_ganguly
```

`docker_container` doesn't have this.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

docker_container

##### ANSIBLE VERSION

```paste below
nsible 2.9.13
  config file = None
  configured module search path = ['/Users/tacyuuhon/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.8.5 (default, Jul 21 2020, 10:48:26) [Clang 11.0.3 (clang-1103.0.32.62)]
```

##### CONFIGURATION

```paste below
ANSIBLE_NOCOWS(/Users/tacyuuhon/ansible/ansible.cfg) = True
HOST_KEY_CHECKING(/Users/tacyuuhon/ansible/ansible.cfg) = False
```

###### OS
MacOS / Darwin w022581902977m 19.6.0 Darwin Kernel Version 19.6.0: Thu Jun 18 20:49:00 PDT 2020; root:xnu-6153.141.1~1/RELEASE_X86_64 x86_64

###### Docker
```
Client: Docker Engine - Community
 Version:           19.03.12
 API version:       1.40
 Go version:        go1.13.10
 Git commit:        48a66213fe
 Built:             Mon Jun 22 15:46:54 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.12
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.10
  Git commit:       48a66213fe
  Built:            Mon Jun 22 15:45:28 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

##### STEPS TO REPRODUCE

```yaml
docker_container:
  name: "nginx"
  image: nginx
  pull: yes
  ports:
    - 8080:8080
    - all
  state: started
```

##### EXPECTED RESULTS

Except for the specified port, the remaining ports are randomly assigned by docker.

##### ACTUAL RESULTS

```paste below
CONTAINER ID        IMAGE               COMMAND                  CREATED              STATUS              PORTS                   NAMES
d1367ad373e0        nginx               "/docker-entrypoint.…"   About a minute ago   Up About a minute   0.0.0.0:32796->80/tcp   nginx
```

Ajpantuso:
##### SUMMARY
Adding `publish_all_ports` option to accurately reflect Docker CLI behavior.
Fixes #8
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME
plugins/modules/docker_container.py
##### ADDITIONAL INFORMATION
The Docker API and Docker CLI combine the use of `ports` and `publish_all_ports` similarly, but the module logic was treating these as mutually exclusive values. This change simply splits the options such that the old way of publishing all ports is available until deprecation, but the new way will work as intended.

The following task:
```yaml
 - hosts: localhost
   collections:
     - community.docker
   tasks:
     - docker_container:
         name: "nginx"
         image: nginx
         pull: yes
         publish_all_ports: true
         ports:
           - 8080:8080
         state: started
       become: true
```
now results in:
```bash
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                                           NAMES
94798f1bf0bf        nginx               "/docker-entrypoint.…"   4 seconds ago       Up 3 seconds        0.0.0.0:8080->8080/tcp, 0.0.0.0:32793->80/tcp   nginx
```

### DIFF
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -663,6 +663,13 @@ options:
       - If I(container_default_behavior) is set to C(compatiblity) (the default value), this
         option has a default of C(no).
     type: bool
+  publish_all_ports:
+    description:
+      - Publish all ports to the host.
+      - Any specified port bindings from I(published_ports) will remain intact when C(true).
+    type: bool
+    default: false
+    version_added: 1.8.0
   published_ports:
     description:
       - List of ports to publish from the container to the host.
@@ -677,7 +684,7 @@ options:
         is different from the C(docker) command line utility. Use the L(dig lookup,../lookup/dig.html)
         to resolve hostnames."
       - A value of C(all) will publish all exposed container ports to random host ports, ignoring
-        any other mappings.
+        any other mappings. Use I(publish_all_ports) instead as the use of C(all) will be deprecated in version 2.0.0.
       - If I(networks) parameter is provided, will inspect each network to see if there exists
         a bridge network with optional parameter C(com.docker.network.bridge.host_binding_ipv4).
         If such a network is found, then published ports where no host IP address is specified
@@ -1424,7 +1431,6 @@ class TaskParameters(DockerBaseClass):
                     except ValueError as exc:
                         self.fail("Failed to convert %s to bytes: %s" % (param_name, to_native(exc)))
 
-        self.publish_all_ports = False
         self.published_ports = self._parse_publish_ports()
         if self.published_ports == 'all':
             self.publish_all_ports = True
@@ -3558,6 +3564,7 @@ def main():
         pid_mode=dict(type='str'),
         pids_limit=dict(type='int'),
         privileged=dict(type='bool'),
+        publish_all_ports=dict(type='bool', default=False),
         published_ports=dict(type='list', elements='str', aliases=['ports']),
         pull=dict(type='bool', default=False),
         purge_networks=dict(type='bool', default=False),
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -1761,7 +1761,7 @@ class TaskParameters(DockerBaseClass):
                     'Specifying "all" in published_ports together with port mappings is not properly '
                     'supported by the module. The port mappings are currently ignored. Set publish_all_ports '
                     'to "true" to randomly assign port mappings for those not specified by published_ports. '
-                    'The use of "all" in published_ports will be removed in version 2.0.0.',
+                    'The use of "all" in published_ports next to other values will be removed in version 2.0.0.',
                     collection_name='community.docker', version='2.0.0')
             return 'all'
 
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -668,7 +668,6 @@ options:
       - Publish all ports to the host.
       - Any specified port bindings from I(published_ports) will remain intact when C(true).
     type: bool
-    default: false
     version_added: 1.8.0
   published_ports:
     description:
@@ -1434,11 +1433,11 @@ class TaskParameters(DockerBaseClass):
         self.published_ports = self._parse_publish_ports()
 
         if self.published_ports == 'all':
-            if not self.publish_all_ports:
+            if self.publish_all_ports is None:
                 self.publish_all_ports = True
                 self.published_ports = None
             else:
-                self.fail('"all" is not a valid value for "published_ports" when "publish_all_ports" is "true"')
+                self.fail('"all" is not a valid value for "published_ports" when "publish_all_ports" is specified')
 
         self.ports = self._parse_exposed_ports(self.published_ports)
         self.log("expose ports:")
@@ -3567,7 +3566,7 @@ def main():
         pid_mode=dict(type='str'),
         pids_limit=dict(type='int'),
         privileged=dict(type='bool'),
-        publish_all_ports=dict(type='bool', default=False),
+        publish_all_ports=dict(type='bool'),
         published_ports=dict(type='list', elements='str', aliases=['ports']),
         pull=dict(type='bool', default=False),
         purge_networks=dict(type='bool', default=False),
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -1432,9 +1432,13 @@ class TaskParameters(DockerBaseClass):
                         self.fail("Failed to convert %s to bytes: %s" % (param_name, to_native(exc)))
 
         self.published_ports = self._parse_publish_ports()
+
         if self.published_ports == 'all':
-            self.publish_all_ports = True
-            self.published_ports = None
+            if not self.publish_all_ports:
+                self.publish_all_ports = True
+                self.published_ports = None
+            else:
+                self.fail('"all" is not a valid value for "published_ports" when "publish_all_ports" is "true"')
 
         self.ports = self._parse_exposed_ports(self.published_ports)
         self.log("expose ports:")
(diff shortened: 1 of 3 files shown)


felixfontein:
You also have to update the deprecation message: https://github.com/ansible-collections/community.docker/blob/18e6bd20d26382ddc94680aecdb3f51df36180a1/plugins/modules/docker_container.py#L1757-L1763.

Also using `all` in `published_ports` should not be allowed if `publish_all_ports` is set to `true`.

Ajpantuso:
> You also have to update the deprecation message:

Done.
> Also using `all` in `published_ports` should not be allowed if `publish_all_ports` is set to `true`.

Module will now fail in this scenario.

felixfontein:
Cool, thanks!

Can you add a test to tests/integration/targets/docker_container/tasks/tests/ports.yml which makes sure that `published_ports: all` is equivalent to `publish_all_ports=true`? (i.e. one more task in the `published_ports: all` series)

plugins/modules/docker_container.py
```diff
@@ -663,6 +663,13 @@
       - If I(container_default_behavior) is set to C(compatiblity) (the default value), this
         option has a default of C(no).
     type: bool
+  publish_all_ports:
+    description:
+      - Publish all ports to the host.
+      - Any specified port bindings from I(published_ports) will remain intact when C(true).
+    type: bool
+    default: false
```
felixfontein, line 671: Right now I'm not really happy with the default for `publish_all_ports`. This is only partially a problem with this PR though, since that behavior was already there before.

The problem is that right now, if the user does not specify `published_ports`, the published ports are not checked for idempotence - **but** the `publish_all_ports` flag is! (If a container has it set, and `docker_container`'s `published_ports` isn't specified, the container will be recreated since there's a config mismatch!) I would argue that the current behavior is a bug, but potentially people depend on it...

The *correct* behavior would be (before this PR) to set `publish_all_ports = None`, and only set it to `True`/`False` if `published_ports` is passed to the module. With the new option it get's a bit more complicated, but it should also not be a problem.

The main problem is what do we do about the old problem. I'm currently tending to treat it as a bug and simply fix it. This changes behavior, but since it prevents container recreation in some situations, I think it's more acceptable than it would be the other way around. What do you think?

Ajpantuso: I tend to agree with you. Any predictable side-effect could be misunderstood as a feature, but the module already has a feature (`recreate`) to let users force recreation in a more obvious way if desired. In other words if they really needed the container to be recreated they should use the intended option otherwise the module is falsely reporting config changes which is a bug.

plugins/modules/docker_container.py
```diff
@@ -1424,11 +1431,14 @@ def __init__(self, client):
                     except ValueError as exc:
                         self.fail("Failed to convert %s to bytes: %s" % (param_name, to_native(exc)))
 
-        self.publish_all_ports = False
         self.published_ports = self._parse_publish_ports()
+
         if self.published_ports == 'all':
-            self.publish_all_ports = True
-            self.published_ports = None
+            if not self.publish_all_ports:
```
felixfontein, line 1437:
```suggestion
            if self.publish_all_ports is None:
```

plugins/modules/docker_container.py
```diff
@@ -1424,11 +1431,14 @@ def __init__(self, client):
...
+                self.publish_all_ports = True
+                self.published_ports = None
+            else:
+                self.fail('"all" is not a valid value for "published_ports" when "publish_all_ports" is "true"')
```
felixfontein, line 1441:
```suggestion
                self.fail('"all" is not a valid value for "published_ports" when "publish_all_ports" is specified')
```

felixfontein (commented):
I think we should also have a set of integration tests for idempotency (and changed status) of `publish_all_ports` combined with `published_ports`.

Ajpantuso:
> I think we should also have a set of integration tests for idempotency (and changed status) of `publish_all_ports` combined with `published_ports`.

Not sure if I understood fully, but I added tests that cover transitions between different combinations of these options.

plugins/modules/docker_container.py
```diff
@@ -1750,10 +1759,9 @@ def _parse_publish_ports(self):
             if len(self.published_ports) > 1:
                 self.client.module.deprecate(
                     'Specifying "all" in published_ports together with port mappings is not properly '
-                    'supported by the module. The port mappings are currently ignored. Please specify '
-                    'only port mappings, or the value "all". The behavior for mixed usage will either '
-                    'be forbidden in version 2.0.0, or properly handled. In any case, the way you '
-                    'currently use the module will change in a breaking way',
+                    'supported by the module. The port mappings are currently ignored. Set publish_all_ports '
+                    'to "true" to randomly assign port mappings for those not specified by published_ports. '
+                    'The use of "all" in published_ports will be removed in version 2.0.0.',
```
felixfontein, line 1764:
```suggestion
                    'The use of "all" in published_ports next to other values will be removed in version 2.0.0.',
```
`all` will still be allowed in 2.0.0. What we will do in 2.0.0 is deprecate `all` (with removal in 3.0.0). But that shouldn't be mentioned here.

tests/integration/targets/docker_container/tasks/tests/ports.yml
```diff
@@ -284,3 +298,154 @@
...
+    force_kill: yes
+  register: publish_all_ports_10
+
+- name: publish_all_ports (false -> null with published_ports)
+  docker_container:
+    image: "{{ docker_test_image_alpine }}"
+    command: '/bin/sh -c "sleep 10m"'
+    name: "{{ cname }}"
+    state: started
+    publish_all_ports: false
```
felixfontein, line 425:
```suggestion
```

felixfontein (commented):
Right now, you're only testing `publish_all_ports` not specified after `publish_all_ports: false`. You must also test `publish_all_ports` not specified after `publish_all_ports: true`, since that's the actually interesting and important case.

Ajpantuso:
> Right now, you're only testing `publish_all_ports` not specified after `publish_all_ports: false`. You must also test `publish_all_ports` not specified after `publish_all_ports: true`, since that's the actually interesting and important case.

Ok, made the tests easier to read this time and 'true to unspecified' is now covered.

tests/integration/targets/docker_container/tasks/tests/ports.yml
```diff
@@ -303,149 +303,78 @@
...
-      - publish_all_ports_4 is not changed
-      - publish_all_ports_5 is not changed
-      - publish_all_ports_6 is changed
-      - publish_all_ports_7 is changed
-      - publish_all_ports_8 is not changed
-      - publish_all_ports_9 is changed
-      - publish_all_ports_10 is not changed
-      - publish_all_ports_11 is not changed
+      - publish_all_ports.results[index].changed == publish_all_ports_test_cases[index].changed
+  loop: "{{ range(0, publish_all_ports_test_cases | length) | list }}"
```
felixfontein, line 378: For things like this, the `indexed_items` lookup is also pretty useful (https://docs.ansible.com/ansible/latest/collections/ansible/builtin/indexed_items_lookup.html#ansible-collections-ansible-builtin-indexed-items-lookup).

felixfontein (approved):

tests/integration/targets/docker_container/tasks/tests/ports.yml
Ajpantuso: Thanks, I updated it to the `loop` analouge.

felixfontein (approved):

felixfontein:
@Ajpantuso thanks a lot for implementing this!

PR #163, opened 2021-06-26
docker_compose - Fix timeout defaulting behavior

### LINKED ISSUE #25: docker_compose module uses inappropriate stop timeout
##### SUMMARY

When the `docker_compose` module is used with `stopped: yes` or `restarted: yes` and no explicit `timeout` argument, it defaults to a 10 second timeout, overriding any `stop_grace_period` that was specified in `docker-compose.yaml`. A more appropriate default would have been to respect `stop_grace_period` unless a timeout was specified explicitly in the ansible task.

Given that this is long-standing behavior, it might not be a good idea to change the default, but perhaps we could introduce a distinguished timeout value that lets the user explicitly specify that the timeout configured in the project should be used.
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

docker_compose
##### ANSIBLE VERSION

```paste below
ansible 2.8.2
```
is what I'm currently using, but the timeout logic hasn't changed since then.
##### CONFIGURATION

```paste below
```

##### OS / ENVIRONMENT

Not relevant

##### STEPS TO REPRODUCE

Create the following three files in $HOME/timeout-bug:

foo.sh:
```bash
#!/usr/bin/env bash

terminate() {
    echo received SIGTERM
    kill -TERM $child
}
echo script has started
trap terminate SIGTERM
/bin/sleep infinity & # stand-in for the real service
child=$!
wait $child
sleep 11 # simulate waiting for a child that takes 11 seconds to finish up
echo exited cleanly
```
docker-compose.yaml:
```yaml
version: "3.2"
services:
  foo:
    image: bash
    command: ["/foo.sh"]
    volumes:
      - type: bind
        source: ./foo.sh
        target: /foo.sh
    stop_grace_period: 12s
```
playbook.yaml:
```yaml
- hosts: localhost
  connection: local
  gather_facts: false
  vars:
    - project_src: "{{ lookup('env', 'HOME') }}/timeout-bug"
  tasks:
    - name: remove pre-existing container to clear logs from previous run
      docker_compose:
        project_src: "{{ project_src }}"
        services:
          - foo
        state: absent

    - name: start service
      docker_compose:
        project_src: "{{ project_src }}"
        services:
          - foo
        state: present

    - name: stop service
      docker_compose:
        project_src: "{{ project_src }}"
        services:
          - foo
        state: present
        stopped: yes
```

Then run the following commands:
```bash
cd $HOME/timeout-bug
ansible-playbook -i localhost, playbook.yaml
docker logs timeoutbug_foo_1
```

##### EXPECTED RESULTS

output should be
```
script has started
received SIGTERM
exited cleanly
```
##### ACTUAL RESULTS

The final line "exited cleanly" is not present in the log, because the container was sent a SIGKILL before it was finished.

```paste below
script has started
received SIGTERM
```

Ajpantuso:
##### SUMMARY
Removed overwriting of default timeout so `stop_grace_period` can be correctly read from `compose` files.
Fixes #25 
##### ISSUE TYPE
- Bugfix Pull Request
##### COMPONENT NAME
plugins/modules/docker_compose.py

##### ADDITIONAL INFORMATION
- `timeout` documentation required `default: null` since a doc fragment this module extends already defines `timeout` with `default: 60`.

After the change the following example from #25 :
```yaml
 tasks:
     - name: remove pre-existing container to clear logs from previous run
       docker_compose:
         project_src: "{{ project_src }}"
         services:
           - foo
         state: absent

     - name: start service
       docker_compose:
         project_src: "{{ project_src }}"
         services:
           - foo
         state: present

     - name: stop service
       docker_compose:
         project_src: "{{ project_src }}"
         services:
           - foo
         state: present
         stopped: yes

     - command: docker logs applications_foo_1
       register: output

     - debug:
         msg: "{{ output.stdout }}"
```
results in:
```bash
PLAY [localhost] ************************************************************************************************************************************************************************************************

TASK [remove pre-existing container to clear logs from previous run] ********************************************************************************************************************************************
changed: [localhost]

TASK [start service] ********************************************************************************************************************************************************************************************
changed: [localhost]

TASK [stop service] *********************************************************************************************************************************************************************************************
changed: [localhost]

TASK [command] **************************************************************************************************************************************************************************************************
changed: [localhost]

TASK [debug] ****************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "script has started\nreceived SIGTERM\nexited cleanly"
}

PLAY RECAP ******************************************************************************************************************************************************************************************************
localhost                  : ok=5    changed=4    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```

### DIFF
--- a/plugins/modules/docker_compose.py
+++ b/plugins/modules/docker_compose.py
@@ -144,8 +144,10 @@ options:
   timeout:
     description:
       - Timeout in seconds for container shutdown when attached or when containers are already running.
+      - By default C(compose) will use a C(10s) timeout unless C(default_grace_period) is defined for a
+        particular service in the I(project_src).
     type: int
-    default: 10
+    default: null
   use_ssh_client:
     description:
       - Currently ignored for this module, but might suddenly be supported later on.
@@ -476,14 +478,13 @@ try:
     from compose.cli.command import project_from_options
     from compose.service import NoSuchImageError
     from compose.cli.main import convergence_strategy_from_opts, build_action_from_opts, image_type_from_opt
-    from compose.const import DEFAULT_TIMEOUT, LABEL_SERVICE, LABEL_PROJECT, LABEL_ONE_OFF
+    from compose.const import LABEL_SERVICE, LABEL_PROJECT, LABEL_ONE_OFF
     HAS_COMPOSE = True
     HAS_COMPOSE_EXC = None
     MINIMUM_COMPOSE_VERSION = '1.7.0'
 except ImportError as dummy:
     HAS_COMPOSE = False
     HAS_COMPOSE_EXC = traceback.format_exc()
-    DEFAULT_TIMEOUT = 10
 
 from ansible.module_utils._text import to_native
 
@@ -1130,7 +1131,7 @@ def main():
         pull=dict(type='bool', default=False),
         nocache=dict(type='bool', default=False),
         debug=dict(type='bool', default=False),
-        timeout=dict(type='int', default=DEFAULT_TIMEOUT)
+        timeout=dict(type='int')
     )
 
     mutually_exclusive = [
--- a/changelogs/fragments/163-docker_compose-timeout-fix.yml
+++ b/changelogs/fragments/163-docker_compose-timeout-fix.yml
@@ -1,4 +1,4 @@
 ---
-bugfixes:
+breaking_changes:
   - docker_compose - fixed ``timeout`` defaulting behavior so that ``stop_grace_period``, if defined in the compose
-    file, will be used if no ``timeout`` is defined (https://github.com/ansible-collections/community.docker/pull/163).
+    file, will be used if `timeout`` is not specified (https://github.com/ansible-collections/community.docker/pull/163).
--- a/changelogs/fragments/163-docker_compose-timeout-fix.yml
+++ b/changelogs/fragments/163-docker_compose-timeout-fix.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - docker_compose - fixed ``timeout`` defaulting behavior so that ``stop_grace_period``, if defined in the compose
+    file, will be used if no ``timeout`` is defined (https://github.com/ansible-collections/community.docker/pull/163).


plugins/modules/docker_compose.py
```diff
@@ -144,8 +144,10 @@
   timeout:
     description:
       - Timeout in seconds for container shutdown when attached or when containers are already running.
+      - By default C(compose) will use a C(10s) timeout unless C(default_grace_period) is defined for a
+        particular service in the I(project_src).
     type: int
-    default: 10
+    default: null
```
felixfontein, line 150:
```suggestion
```

felixfontein (commented):
Isn't this a breaking change?

Ajpantuso:
> Isn't this a breaking change?

I detailed why it wasn't on the [issue thread](https://github.com/ansible-collections/community.docker/issues/25#issuecomment-868903101)
Basically the module was enforcing a timeout before `compose` could apply it's defaulting logic and it was doing this by importing the `DEFAULT_TIMEOUT` setting from `compose`. So letting `compose` handle the defaulting itself results in the same timeout settings. I guess it's only breaking if this module never intended to honor `stop_grace_period`.

felixfontein:
Well, it's still a breaking change, since in some cases (user had a different timeout in the service description than the default value) the module will change behavior. I tend to agree with @aaron-kaplan on:

> Given that this is long-standing behavior, it might not be a good idea to change the default, but perhaps we could introduce a distinguished timeout value that lets the user explicitly specify that the timeout configured in the project should be used.

I think we should probably keep this for the 2.0.0 release and add a `breaking_change` changelog fragment. That way, if someone really needs the old behavior, they won't get surprised in a minor release.

Ajpantuso:
> Well, it's still a breaking change, since in some cases (user had a different timeout in the service description than the default value) the module will change behavior. I tend to agree with @aaron-kaplan on:
> 
> > Given that this is long-standing behavior, it might not be a good idea to change the default, but perhaps we could introduce a distinguished timeout value that lets the user explicitly specify that the timeout configured in the project should be used.
> 
> I think we should probably keep this for the 2.0.0 release and add a `breaking_change` changelog fragment. That way, if someone really needs the old behavior, they won't get surprised in a minor release.

Updated the changelog fragment.
On a side note I was going to implement `profiles` for `docker_compose` since that's only a few lines of implementation (and the CLI option is bugged such that it only works when bringing services up so not much to test), but I can hold that for after 1.8.0 if you don't want to wait.

felixfontein:
> Updated the changelog fragment.

Thanks! I guess we'll start with the 2.0.0 preparations somewhen in September; I'll merge this PR then.

> On a side note I was going to implement `profiles` for `docker_compose` since that's only a few lines of implementation (and the CLI option is bugged such that it only works when bringing services up so not much to test), but I can hold that for after 1.8.0 if you don't want to wait.

Feel free to start working on that now if you want. If the PR is done merged before I do the 1.8.0 release (probably on Tuesday morning), it will be in 1.8.0, otherwise it has to wait for 1.9.0.

felixfontein:
Restarting tests.

felixfontein:
@Ajpantuso thanks for contributing this!

Issue #24, opened 2020-05-16
[Feature Request] docker-compose config -q


lonix1:
I was told to move this here, from https://github.com/ansible/ansible/issues/62745

---

A `docker-compose` file (and a corresponding `.env` file), can be validated for correctness like this:

````
docker-compose config -q
````

Please add that capability to the [docker_compose](https://docs.ansible.com/ansible/latest/modules/docker_compose_module.html) module.

Many modules have a `validate` argument, perhaps the same can be done here?

felixfontein:
The community.docker.docker_compose module has been removed from today's 4.0.0 release. If you are still using it, please upgrade to community.docker.docker_compose_v2, which no longer uses the End of Life versions Docker Compose 1.x.y, but Docker's Compose plugin (version 2.x.y range).

PR #169, opened 2021-06-29
docker-compose - Adding validated state and no_interpolate option

Ajpantuso:
##### SUMMARY
Adds a new state `validated` along with a new option `no_interpolate` which is the equivalent of `docker-compose config`
Implements #24
##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME
plugins/modules/docker-compose.py
##### ADDITIONAL INFORMATION
- Original request was just for validating configs with no output which is already covered by `check_mode` so the `validated` state was extended to actually display output like `docker-compose config`.
- `no-interpolate` adds the ability to show unsubstituted environment variables in the `config` output.

### DIFF
--- a/162-docker_container_publish_all_option.yml
+++ b/162-docker_container_publish_all_option.yml
@@ -1,4 +1,0 @@
----
-minor_changes:
-  - docker_container - added ``publish_all_ports`` option to publish all exposed ports to random ports except those
-    explicitly bound with ``published_ports`` (https://github.com/ansible-collections/community.docker/pull/162).
--- a/plugins/modules/docker_compose.py
+++ b/plugins/modules/docker_compose.py
@@ -56,11 +56,14 @@ options:
       - Specifying C(present) is the same as running C(docker-compose up) resp. C(docker-compose stop) (with I(stopped)) resp. C(docker-compose restart)
         (with I(restarted)).
       - Specifying C(absent) is the same as running C(docker-compose down).
+      - Specifying C(validated) is the same as running C(docker-compose config) which validates the provided compose
+        files and returns the contained configuration.
     type: str
     default: present
     choices:
       - absent
       - present
+      - validated
   services:
     description:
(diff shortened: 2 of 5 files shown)


162-docker_container_publish_all_option.yml
```diff
@@ -1,4 +0,0 @@
----
-minor_changes:
```
felixfontein, line 2: Can you rebase to latest `main` so these 'fake' changes go away? They're making reviewing harder :)

plugins/modules/docker_compose.py
```diff
@@ -455,6 +464,12 @@
                           description: the container's short ID
                           returned: always
                           type: str
+config:
+  description:
+    - The resulting configuration from one or more compose files.
+    - Equivalent to the output of C(docker-compose config).
+  returned: when success and I(state=validated)
+  type: dict
```
felixfontein, line 472:
```suggestion
  type: dict
  version_added: 1.9.0
```

felixfontein (commented):
I'm not sure that we should add this as a state. I think it would be better to have this as an `_info` module; maybe `docker_compose_validate_info`?

162-docker_container_publish_all_option.yml
Ajpantuso: This is a real change. I missed moving this to `changelogs/fragments` in #162 so now it's just hanging out in root after 1.8.0.

Ajpantuso:
> I'm not sure that we should add this as a state. I think it would be better to have this as an `_info` module; maybe `docker_compose_validate_info`?

Yeah, I also considered this. If I go that route then I guess it could be `docker_compose_config_info` and just implement the features of [docker-compose config}(https://docs.docker.com/compose/reference/config/).

162-docker_container_publish_all_option.yml
felixfontein: Ooops. I completely missed that...

I'll do the move in a separate commit (and add a note that this was already added in 1.8.0).

Ajpantuso: Thanks, I'll close this and look into an info module instead.

PR #170, opened 2021-07-01
put example addresses in actual RFC1918 range

lotheac:
172.1/16, 172.3/16 and 172.4/16 are not in the 172.16/12 range.

https://datatracker.ietf.org/doc/html/rfc1918#section-3

##### SUMMARY
The docker_network & docker_container examples threw me off since they were using non-private IPv4 address spaces. This seems like a mistake, so replace them with actual private address spaces.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
docker_container
docker_network

### DIFF
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -981,7 +981,7 @@ EXAMPLES = '''
     image: ubuntu:14.04
     networks:
       - name: TestingNet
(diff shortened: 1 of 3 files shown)


felixfontein (approved):

felixfontein:
@lotheac thanks for this improvement!

PR #173, opened 2021-07-14
Add nsenter connection plugin

jgoldschrafe:
##### SUMMARY
Add a new `nsenter` connection plugin, which uses the `nsenter` utility to run commands in the context of another namespace.

Currently, if you are running Ansible in a pull capacity (i.e. a host downloads Ansible content and executes plays against itself), it is difficult to run the Ansible controller in a Docker container; the `local` connection type is not effective as the container is an isolated, ephemeral environment. If you are very insistent, you may do something like grant the container access to the host OS over SSH. This can be complicated and error-prone, especially with complex Docker networking configurations.

This connection plugin allows a containerized Ansible controller, if launched in a specific way, to escape the container (if it is granted those privileges) and manage the host it is running on, dramatically simplifying the process of managing Ansible versions and their dependencies, especially for hosts running Python < 3.8 (such as Ubuntu 18.04).

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
nsenter

##### ADDITIONAL INFORMATION
The plugin is largely based on the `local` connection plugin from Ansible core, with some small changes.

An example invocation of the container might look like:

```
docker run \
    --rm \
    --privileged \
    --pid host \
    --volume /:/host \
    --volume /my/ansible/content:/my/ansible/content \
    --workdir /my/ansible/content \
    my-ansible-image:latest \
    ansible-playbook playbooks/my-playbook.yml
```

It is hypothetically possible for this to work running as a non-root user in the container. It has not been tested.

The test suite is currently marked as `unsupported` because invocation is a mess and `runme.sh` directly invokes a container to run ansible-playbook. `ansible-test` allows suites to be marked as privileged via the `needs/privileged` integration alias, but does not currently support namespace options (i.e. `--pid host`) or volume mounts.

### DIFF
--- a/plugins/connection/nsenter.py
+++ b/plugins/connection/nsenter.py
@@ -234,7 +234,7 @@ class Connection(ConnectionBase):
             rc, out, err = self.exec_command(cmd=["cat", in_path])
             display.vvv(u"FETCH {0} TO {1}".format(in_path, out_path), host=self._play_context.remote_addr)
             if rc != 0:
-                raise AnsibleError('Failed to fetch file `{0}`: {1}'.format(in_path, err))
+                raise AnsibleError("failed to transfer file to {0}: {1}".format(in_path, err))
             with open(to_bytes(out_path, errors='surrogate_or_strict'), 'wb') as out_file:
                 out_file.write(out)
         except IOError as e:
--- a/plugins/connection/nsenter.py
+++ b/plugins/connection/nsenter.py
@@ -1,0 +1,250 @@
+# (c) 2021 Jeff Goldschrafe <jeff@holyhandgrenade.org>
+# Based on Ansible local connection plugin by:
+# (c) 2012 Michael DeHaan <michael.dehaan@gmail.com>
+# (c) 2015, 2017 Toshio Kuratomi <tkuratomi@ansible.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = '''
+    name: nsenter
+    short_description: execute on host running controller container
+    description:
+        - This connection plugin allows Ansible, running in a privileged container, to execute tasks on the container
+          host instead of in the container itself.
+        - This is useful for running Ansible in a pull model, while still keeping the Ansible control node
+          containerized.
+        - It relies on having privileged access to run nsenter in the host's PID namespace, allowing it to enter the
+          namespace of PID 1.
+    author: Jeff Goldschrafe (@jgoldschrafe)
+    options:
+        host_volume_mount:
+            description:
+                - Host volume mount to read and write host files through
+            default: /host
+            vars:
+                - name: ansible_host_volume_mount
+            env:
+                - name: ANSIBLE_HOST_VOLUME_MOUNT
+            ini:
+                - section: nsenter_connection
+                  key: host_volume_mount
+        nsenter_pid:
+            description:
+                - PID to attach with using nsenter.
+                - The default should be fine unless you're attaching as a non-root user.
+            default: '1'
+            vars:
+                - name: ansible_nsenter_pid
+            env:
+                - name: ANSIBLE_NSENTER_PID
+            ini:
+                - section: nsenter_connection
+                  key: nsenter_pid
+        pipelining:
+            default: ANSIBLE_PIPELINING
+            description:
+                - Pipelining reduces the number of connection operations required to execute a module on the remote
+                  server, by executing many Ansible modules without actual file transfers.
+                - This can result in a very significant performance improvement when enabled.
+                - However this can conflict with privilege escalation (become).
+                  For example, when using sudo operations you must first disable 'requiretty' in the sudoers file for
+                  the target hosts, which is why this feature is disabled by default.
+            env:
+                - name: ANSIBLE_PIPELINING
+            ini:
+                - section: defaults
+                  key: pipelining
+            type: boolean
+            vars:
+                - name: ansible_pipelining
+    notes:
+        - The remote user is ignored; this plugin always runs as root.
+        - This plugin requires the container to be launched in the following way:
+            - The container image contains the C(nsenter) program
+            - C(/) on the host is mounted read-write to C(/host) in the container
+            - The container is launched in privileged mode
+            - The container is launched in the host's PID namespace using C(--pid host)
+'''
+
+import os
+import pty
+import shutil
+import subprocess
+import fcntl
+
+import ansible.constants as C
+from ansible.errors import AnsibleError, AnsibleFileNotFound
+from ansible.module_utils.compat import selectors
+from ansible.module_utils.six import text_type, binary_type
+from ansible.module_utils._text import to_bytes, to_native, to_text
+from ansible.plugins.connection import ConnectionBase
+from ansible.utils.display import Display
+from ansible.utils.path import unfrackpath
+
+display = Display()
+
+
+class Connection(ConnectionBase):
+    '''Connections to a container host using nsenter
+    '''
+
+    transport = 'nsenter'
+    has_pipelining = True
+
+    def __init__(self, *args, **kwargs):
+        super(Connection, self).__init__(*args, **kwargs)
+        self.cwd = None
+        self._host_volume_mount = self.get_option("host_volume_mount")
+        self._nsenter_pid = self.get_option("nsenter_pid")
+
+        if not os.path.exists(self._host_volume_mount):
+            raise AnsibleError("nsenter: host volume mount does not exist: " + self._host_volume_mount)
+
+    def _connect(self):
+        # Because nsenter requires very high privileges, our remote user
+        # is always assumed to be root.
+        self._play_context.remote_user = "root"
+
+        if not self._connected:
+            display.vvv(
+                u"ESTABLISH NSENTER CONNECTION FOR USER: {0}".format(
+                    self._play_context.remote_user
+                ),
+                host=self._play_context.remote_addr,
+            )
+            self._connected = True
+        return self
+
+    def exec_command(self, cmd, in_data=None, sudoable=True):
+        super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
+
+        display.debug("in nsenter.exec_command()")
+
+        executable = C.DEFAULT_EXECUTABLE.split()[0] if C.DEFAULT_EXECUTABLE else None
+
+        if not os.path.exists(to_bytes(executable, errors='surrogate_or_strict')):
+            raise AnsibleError("failed to find the executable specified %s."
+                               " Please verify if the executable exists and re-try." % executable)
+
+        # Rewrite the provided command to prefix it with nsenter
+        if isinstance(cmd, (text_type, binary_type)):
+            nsenter_cmd = "nsenter --all --preserve-credentials --target={0} -- ".format(self._nsenter_pid)
+            cmd = to_bytes(nsenter_cmd + cmd)
+        else:
+            nsenter_cmd = ["nsenter", "--all", "--preserve-credentials", "--target", self._nsenter_pid, "--"]
+            cmd = map(to_bytes, nsenter_cmd + cmd)
+
+        display.vvv(u"EXEC {0}".format(to_text(cmd)), host=self._play_context.remote_addr)
+        display.debug("opening command with Popen()")
+
+        master = None
+        stdin = subprocess.PIPE
+        if sudoable and self.become and self.become.expect_prompt() and not self.get_option('pipelining'):
+            # Create a pty if sudoable for privlege escalation that needs it.
+            # Falls back to using a standard pipe if this fails, which may
+            # cause the command to fail in certain situations where we are escalating
+            # privileges or the command otherwise needs a pty.
+            try:
+                master, stdin = pty.openpty()
+            except (IOError, OSError) as e:
+                display.debug("Unable to open pty: %s" % to_native(e))
+
+        p = subprocess.Popen(
+            cmd,
+            shell=isinstance(cmd, (text_type, binary_type)),
+            executable=executable,
+            cwd=self.cwd,
+            stdin=stdin,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+
+        # if we created a master, we can close the other half of the pty now, otherwise master is stdin
+        if master is not None:
+            os.close(stdin)
+
+        display.debug("done running command with Popen()")
+
+        if self.become and self.become.expect_prompt() and sudoable:
+            fcntl.fcntl(p.stdout, fcntl.F_SETFL, fcntl.fcntl(p.stdout, fcntl.F_GETFL) | os.O_NONBLOCK)
+            fcntl.fcntl(p.stderr, fcntl.F_SETFL, fcntl.fcntl(p.stderr, fcntl.F_GETFL) | os.O_NONBLOCK)
+            selector = selectors.DefaultSelector()
+            selector.register(p.stdout, selectors.EVENT_READ)
+            selector.register(p.stderr, selectors.EVENT_READ)
+
+            become_output = b''
+            try:
+                while not self.become.check_success(become_output) and not self.become.check_password_prompt(become_output):
+                    events = selector.select(self._play_context.timeout)
+                    if not events:
+                        stdout, stderr = p.communicate()
+                        raise AnsibleError('timeout waiting for privilege escalation password prompt:\n' + to_native(become_output))
+
+                    for key, event in events:
+                        if key.fileobj == p.stdout:
+                            chunk = p.stdout.read()
+                        elif key.fileobj == p.stderr:
+                            chunk = p.stderr.read()
+
+                    if not chunk:
+                        stdout, stderr = p.communicate()
+                        raise AnsibleError('privilege output closed while waiting for password prompt:\n' + to_native(become_output))
+                    become_output += chunk
+            finally:
+                selector.close()
+
+            if not self.become.check_success(become_output):
+                become_pass = self.become.get_option('become_pass', playcontext=self._play_context)
+                if master is None:
+                    p.stdin.write(to_bytes(become_pass, errors='surrogate_or_strict') + b'\n')
+                else:
+                    os.write(master, to_bytes(become_pass, errors='surrogate_or_strict') + b'\n')
+
+            fcntl.fcntl(p.stdout, fcntl.F_SETFL, fcntl.fcntl(p.stdout, fcntl.F_GETFL) & ~os.O_NONBLOCK)
+            fcntl.fcntl(p.stderr, fcntl.F_SETFL, fcntl.fcntl(p.stderr, fcntl.F_GETFL) & ~os.O_NONBLOCK)
+
+        display.debug("getting output with communicate()")
+        stdout, stderr = p.communicate(in_data)
+        display.debug("done communicating")
+
+        # finally, close the other half of the pty, if it was created
+        if master:
+            os.close(master)
+
+        display.debug("done with nsenter.exec_command()")
+        return (p.returncode, stdout, stderr)
+
+    def put_file(self, in_path, out_path):
+        super(Connection, self).put_file(in_path, out_path)
+
+        # Rewrite out_path into host volume mount
+        in_path = unfrackpath(in_path, basedir=self.cwd)
+        out_path = self._host_volume_mount + unfrackpath(out_path, basedir=self.cwd)
+
+        return self._transfer_file("fetch", in_path, out_path)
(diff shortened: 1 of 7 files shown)


plugins/connection/nsenter.py
```diff
@@ -0,0 +1,250 @@
...
+# (c) 2012 Michael DeHaan <michael.dehaan@gmail.com>
+# (c) 2015, 2017 Toshio Kuratomi <tkuratomi@ansible.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = '''
+    name: nsenter
+    short_description: execute on host running controller container
```
felixfontein, line 12:
```suggestion
    short_description: execute on host running controller container
    version_added: 1.9.0
```

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,250 @@
...
+          host instead of in the container itself.
+        - This is useful for running Ansible in a pull model, while still keeping the Ansible control node
+          containerized.
+        - It relies on having privileged access to run nsenter in the host's PID namespace, allowing it to enter the
+          namespace of PID 1.
+    author: Jeff Goldschrafe (@jgoldschrafe)
+    options:
+        host_volume_mount:
+            description:
+                - Host volume mount to read and write host files through
```
felixfontein, line 24:
```suggestion
                - Host volume mount to read and write host files through.
```
Descriptions should always have complete sentences.

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,250 @@
...
+            ini:
+                - section: defaults
+                  key: pipelining
+            type: boolean
+            vars:
+                - name: ansible_pipelining
+    notes:
+        - The remote user is ignored; this plugin always runs as root.
+        - This plugin requires the container to be launched in the following way:
+            - The container image contains the C(nsenter) program
```
felixfontein, line 65: Having an enumeration inside a notes paragraph doesn't work (i.e. it will look horribly). Try to format this as one paragraph, like "Bla (1) foo, (2) bar, (3) baz."

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,250 @@
...
+        - This is useful for running Ansible in a pull model, while still keeping the Ansible control node
+          containerized.
+        - It relies on having privileged access to run nsenter in the host's PID namespace, allowing it to enter the
+          namespace of PID 1.
+    author: Jeff Goldschrafe (@jgoldschrafe)
+    options:
+        host_volume_mount:
+            description:
+                - Host volume mount to read and write host files through
+            default: /host
```
felixfontein, line 25:
```suggestion
            type: string
            default: /host
```
Please always specify `type`.

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,250 @@
...
+            env:
+                - name: ANSIBLE_HOST_VOLUME_MOUNT
+            ini:
+                - section: nsenter_connection
+                  key: host_volume_mount
+        nsenter_pid:
+            description:
+                - PID to attach with using nsenter.
+                - The default should be fine unless you're attaching as a non-root user.
+            default: '1'
```
felixfontein, line 37:
```suggestion
            type: int
            default: 1
```
Shouldn't this be an integer and not a string?

felixfontein (commented):
Thanks for your contribution! I've added some first comments; I'll try to take a closer look this evening.

Please note that the sanity tests are failing (there's a newline missing at the end of the file, and `ansible-doc` currently has some problem with your plugin).

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,259 @@
...
+    options:
+        host_volume_mount:
+            description:
+                - Host volume mount to read and write host files through.
+            type: string
+            default: /host
+            vars:
+                - name: ansible_host_volume_mount
+            env:
+                - name: ANSIBLE_HOST_VOLUME_MOUNT
```
briantist, line 31: I might recommend adding nsenter into these names.
```suggestion
            vars:
                - name: ansible_nsenter_host_volume_mount
            env:
                - name: ANSIBLE_NSENTER_HOST_VOLUME_MOUNT
```

jgoldschrafe: One better: I funneled the file I/O through `nsenter` as well, and made the volume mounts unnecessary. Nice flag, thank you!

felixfontein:
@jgoldschrafe can you please remove the `@` from the commit message? Having a `@` in there gets me one notification (which results in an email for me) every time someone pushes that commit somewhere :)

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,245 @@
...
+            ini:
+                - section: defaults
+                  key: pipelining
+            type: boolean
+            vars:
+                - name: ansible_pipelining
+    notes:
+        - The remote user is ignored; this plugin always runs as root.
+        - "This plugin requires the Ansible controller container to be launched in the following way:"
+        - (1) The container image contains the nsenter program;
```
felixfontein, line 56: Unfortunately this won't work either (at least not long-term), since once you include notes from a doc fragment, Ansible will sort the notes alphabetically, and these will end up at the top of the list without context :)

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,245 @@
...
+from ansible.utils.path import unfrackpath
+
+display = Display()
+
+
+class Connection(ConnectionBase):
+    '''Connections to a container host using nsenter
+    '''
+
+    transport = 'nsenter'
```
felixfontein, line 83:
```suggestion
    transport = 'community.docker.nsenter'
```

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,245 @@
...
+import os
+import pty
+import shutil
+import subprocess
+import fcntl
+
+import ansible.constants as C
+from ansible.errors import AnsibleError, AnsibleFileNotFound
+from ansible.module_utils.compat import selectors
+from ansible.module_utils.six import text_type, binary_type
```
felixfontein, line 70:
```suggestion
from ansible.module_utils.six import string_types
```

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,245 @@
...
+        display.debug("in nsenter.exec_command()")
+
+        executable = C.DEFAULT_EXECUTABLE.split()[0] if C.DEFAULT_EXECUTABLE else None
+
+        if not os.path.exists(to_bytes(executable, errors='surrogate_or_strict')):
+            raise AnsibleError("failed to find the executable specified %s."
+                               " Please verify if the executable exists and re-try." % executable)
+
+        # Rewrite the provided command to prefix it with nsenter
+        if isinstance(cmd, (text_type, binary_type)):
```
felixfontein, line 118:
```suggestion
        if isinstance(cmd, string_types):
```

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,245 @@
...
+            nsenter_cmd = "nsenter --all --preserve-credentials --target={0} -- ".format(self._nsenter_pid)
+            cmd = to_bytes(nsenter_cmd + cmd)
```
felixfontein, line 120:
```suggestion
            cmd = to_bytes(nsenter_cmd) + to_bytes(cmd)
```
Otherwise you can end up with funny string contents when `nsenter_cmd` and `cmd` are not of the same type.

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,245 @@
...
+        else:
+            nsenter_cmd = [
+                "nsenter",
+                "--all",
+                "--preserve-credentials",
+                "--target=" + str(self._nsenter_pid),
```
felixfontein, line 126:
```suggestion
                "--target={0}".format(self._nsenter_pid),
```

felixfontein:
@jgoldschrafe ping

jgoldschrafe:
I fixed the docstring, but reverted the `(text_type, binary_type)` -> `string_types` change because it broke construction of the command line. That code is largely copied out of the `local` connection plugin, so I'd prefer to track it directly anyway except where there's a good reason to diverge.

tests/integration/targets/connection_nsenter/aliases
```diff
@@ -0,0 +1,3 @@
+skip/docker  # this requires unfettered access to the container host
+destructive
+unsupported
```
felixfontein, line 3: How do you know that the tests do not work in CI?

briantist: I'm not certain but I'd guess it's related to these requirements from the notes:
```
            This plugin requires the Ansible controller container to be launched in the following way:
            (1) The container image contains the nsenter program;
            (2) The container is launched in privileged mode;
            (3) The container is launched in the host's PID namespace (i.e. C(--pid host)).
```

felixfontein: The tests are launching the container, so for the CI instances which are VMs and not containers this should just work. (For the CI instances which are not VMs the tests are alread skipped by the `skip/docker` above.)

briantist: neat, I bet it's worth a try then!

felixfontein, line 3:
```suggestion
shippable/posix/group5
skip/docker  # this requires unfettered access to the container host
destructive
```

felixfontein:
Let's see what happens :)

briantist:
🍿👀

tests/integration/targets/connection_nsenter/runme.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+[[ -n "${DEBUG:-}" || -n "${ANSIBLE_DEBUG:-}" ]] && set -x
+
+readonly IMAGE="quay.io/ansible/toolset:latest"
+readonly COLLECTION_ROOT="$(git rev-parse --show-toplevel)"
```
felixfontein, line 8: Neither is `git` installed, nor is there always a git repository around when tests are run.
```suggestion
readonly COLLECTION_ROOT="$(cd ../../../.. ; pwd)"
```

tests/integration/targets/connection_nsenter/runme.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+[[ -n "${DEBUG:-}" || -n "${ANSIBLE_DEBUG:-}" ]] && set -x
+
+readonly IMAGE="quay.io/ansible/toolset:latest"
+readonly COLLECTION_ROOT="$(cd ../../../.. ; pwd)"
```
felixfontein, line 8:
```suggestion
readonly COLLECTION_ROOT="$(cd ../../../.. ; pwd)"
readonly COLLECTIONS_PATH="$(cd ../../../../../../.. ; pwd)"
```

tests/integration/targets/connection_nsenter/runme.sh
```diff
@@ -0,0 +1,61 @@
...
+}
+
+envs=(--env "HOME=${HOME:-}")
+while IFS=$'\0' read -d '' -r line; do
+    key="$(echo "$line" | cut -d= -f1)"
+    value="$(echo "$line" | cut -d= -f2-)"
+    if [[ "${key}" =~ ^(ANSIBLE_|JUNIT_OUTPUT_DIR$|OUTPUT_DIR$|PYTHONPATH$) ]]; then
+        envs+=(--env "${key}=${value}")
+    fi
+done < <(printenv -0)
```
felixfontein, line 35:
```suggestion
done < <(printenv -0)

# Make sure the directory containing ansible_collections is in Ansible's search path
if [ "${ANSIBLE_COLLECTIONS_PATHS}" == "" ]; then
    envs+=(--env "ANSIBLE_COLLECTIONS_PATHS=${COLLECTIONS_PATH}")
else
    envs+=(--env "ANSIBLE_COLLECTIONS_PATHS=${COLLECTIONS_PATH}:${ANSIBLE_COLLECTIONS_PATHS}")
fi
```

tests/integration/targets/connection_nsenter/runme.sh
```diff
@@ -0,0 +1,61 @@
...
+
+[nsenter:vars]
+ansible_host=localhost
+ansible_connection=community.docker.nsenter
+ansible_host_volume_mount=/host
+ansible_nsenter_pid=1
+ansible_python_interpreter=${PYTHON}
+EOF
+
+echo "Run tests"
```
felixfontein, line 59:
```suggestion
echo "Run tests"
set -x
```

tests/integration/targets/connection_nsenter/aliases
felixfontein: No, without the correct target the tests will never be run.

tests/integration/targets/connection_nsenter/runme.sh
```diff
@@ -0,0 +1,70 @@
...
+        envs+=(--env "${key}=${value}")
+    fi
+done < <(printenv -0)
+
+# Make sure the directory containing ansible_collections is in Ansible's search path
+if [ "${ANSIBLE_COLLECTIONS_PATHS}" == "" ]; then
+    envs+=(--env "ANSIBLE_COLLECTIONS_PATHS=${COLLECTIONS_PATH}")
+else
+    envs+=(--env "ANSIBLE_COLLECTIONS_PATHS=${COLLECTIONS_PATH}:${ANSIBLE_COLLECTIONS_PATHS}")
+fi
```
felixfontein, line 43:
```suggestion
if [ -v ANSIBLE_COLLECTIONS_PATHS ]; then
    envs+=(--env "ANSIBLE_COLLECTIONS_PATHS=${COLLECTIONS_PATH}:${ANSIBLE_COLLECTIONS_PATHS}")
else
    envs+=(--env "ANSIBLE_COLLECTIONS_PATHS=${COLLECTIONS_PATH}")
fi
```

tests/integration/targets/connection_nsenter/runme.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+[[ -n "${DEBUG:-}" || -n "${ANSIBLE_DEBUG:-}" ]] && set -x
+
+readonly IMAGE="quay.io/ansible/toolset:latest"
+readonly COLLECTION_ROOT="$(cd ../../../.. ; pwd)"
+readonly PYTHON="$(command -v python3 python | head -n1)"
```
felixfontein, line 9:
```suggestion
readonly PYTHON="$(command -v python3 python | head -n1)"

# Determine collection root
COLLECTION_ROOT=./
while true; do
    if [ -e ${COLLECTION_ROOT}galaxy.yml ] || [ -e ${COLLECTION_ROOT}MANIFEST.json ]; then
        break
    fi
    COLLECTION_ROOT="${COLLECTION_ROOT}../"
done
readonly COLLECTION_ROOT="$(cd ${COLLECTION_ROOT} ; pwd)"
```

tests/integration/targets/connection_nsenter/aliases
```diff
@@ -0,0 +1,3 @@
+shippable/posix/group5
+skip/docker  # this requires unfettered access to the container host
```
felixfontein, line 2:
```suggestion
skip/docker  # this requires unfettered access to the container host
skip/rhel7.9  # nsenter does not work out of the box
```

felixfontein:
CI now passes, you can see logs of the two actual CI runs here:
- https://dev.azure.com/ansible/22c50118-fc97-4e5a-9c0d-36527697c9bb/_apis/build/builds/21039/logs/755
- https://dev.azure.com/ansible/22c50118-fc97-4e5a-9c0d-36527697c9bb/_apis/build/builds/21039/logs/757

I'm not completely happy about the usage of quay.io/ansible/toolset, since that image is pretty large and contains another ansible-core version than the one the CI is using, but for now this is a lot better than no tests at all.

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,246 @@
...
+                - section: defaults
+                  key: pipelining
+            type: boolean
+            vars:
+                - name: ansible_pipelining
+    notes:
+        - The remote user is ignored; this plugin always runs as root.
+        - >-
+            This plugin requires the Ansible controller container to be launched in the following way:
+            (1) The container image contains the nsenter program;
```
felixfontein, line 57:
```suggestion
            (1) The container image contains the C(nsenter) program;
```

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,246 @@
...
+            (2) The container is launched in privileged mode;
+            (3) The container is launched in the host's PID namespace (i.e. C(--pid host)).
```
felixfontein, line 59:
```suggestion
            (3) The container is launched in the host's PID namespace (C(--pid host)).
```

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,246 @@
...
+import pty
+import shutil
+import subprocess
+import fcntl
+
+import ansible.constants as C
+from ansible.errors import AnsibleError, AnsibleFileNotFound
+from ansible.module_utils.compat import selectors
+from ansible.module_utils.six import binary_type, text_type
+from ansible.module_utils._text import to_bytes, to_native, to_text
```
felixfontein, line 72:
```suggestion
from ansible.module_utils.common.text.converters import to_bytes, to_native, to_text
```

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,246 @@
...
+DOCUMENTATION = '''
+    name: nsenter
+    short_description: execute on host running controller container
+    version_added: 1.9.0
+    description:
+        - This connection plugin allows Ansible, running in a privileged container, to execute tasks on the container
+          host instead of in the container itself.
+        - This is useful for running Ansible in a pull model, while still keeping the Ansible control node
+          containerized.
+        - It relies on having privileged access to run nsenter in the host's PID namespace, allowing it to enter the
```
felixfontein, line 19:
```suggestion
        - It relies on having privileged access to run C(nsenter) in the host's PID namespace, allowing it to enter the
```

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,246 @@
...
+          namespaces of the provided PID (default PID 1, or init/systemd).
+    author: Jeff Goldschrafe (@jgoldschrafe)
+    options:
+        nsenter_pid:
+            description:
+                - PID to attach with using nsenter.
+                - The default should be fine unless you're attaching as a non-root user.
```
felixfontein, line 26:
```suggestion
                - The default should be fine unless you are attaching as a non-root user.
```

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,246 @@
...
+            type: int
+            default: 1
+            vars:
+                - name: ansible_nsenter_pid
+            env:
+                - name: ANSIBLE_NSENTER_PID
+            ini:
+                - section: nsenter_connection
+                  key: nsenter_pid
+        pipelining:
```
felixfontein, line 36:
```suggestion
        pipelining:
            # This is part of the ansible-core 2.12 docs fragment 'connection_pipelining'.
```

plugins/connection/nsenter.py
```diff
@@ -0,0 +1,246 @@
...
+            default: ANSIBLE_PIPELINING
```
felixfontein, line 37: That's a string, not a boolean. The default must be `true` or `false`.

(Personally I would remove this whole option and do what the `local` plugin has been doing until ansible-core 2.11, namely ignore it.)

briantist: if it were removed I might recommend keeping at least part of it (the name?) around, and commented out, with a note about why, just so that if the parts from `local` are "refreshed" it doesn't get brought back in accidentally.

plugins/connection/nsenter.py
briantist: Not that everything everywhere even remotely follows this, but just pointing out (as I only saw it recently myself) that there is an Ansible style guide and they directly address use of Latin 😅:
https://docs.ansible.com/ansible/devel/dev_guide/style_guide/#avoid-using-latin-phrases
(some of the 🙅‍♂️ things in this guide are things I do a lot haha...)

plugins/connection/nsenter.py
felixfontein: For that a remark in line https://github.com/ansible-collections/community.docker/pull/173/files#diff-fd1d8cb72d9ad37889af492d0e7412b05cb16311b91327493e556af76b7ba645R137 should be enough.

jgoldschrafe:
Updated again - thank you both for the excellent feedback and advice!

felixfontein (approved):

felixfontein:
@jgoldschrafe thanks for contributing this connection plugin!
@briantist thanks for reviewing!

PR #179, opened 2021-07-26
Add default value if 'registry_url' is provided and its empty

keducoop:
##### SUMMARY

Add default value if 'registry_url' is provided and its empty.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
docker_login

##### ADDITIONAL INFORMATION
    
Currently If 'registry_url' is provided and its an empty string:

```
- name: Log into default docker registry
  docker_login:
    registry: 
    username: my_good_username
    password: my_good_temporary_token
    reauthorize: yes
  become: true
```

 it will cause below `~/.docker/config.json`:
    
```
     {
        "auths": {
            "": {
                "auth": "my_good_temporary_token"
            }
        }
    }
```

Which will prevent ulterior commands such as `docker pull`.

### DIFF
--- a/plugins/modules/docker_login.py
+++ b/plugins/modules/docker_login.py
@@ -288,7 +288,7 @@ class LoginManager(DockerBaseClass):
         parameters = self.client.module.params
         self.check_mode = self.client.check_mode
 
-        self.registry_url = parameters.get('registry_url')
+        self.registry_url = parameters.get('registry_url') or 'https://index.docker.io/v1/'
         self.username = parameters.get('username')
         self.password = parameters.get('password')
         self.email = parameters.get('email')


felixfontein:
If you provide that option with an empty string, that is exactly what you ask the module to do. I don't see why this should suddenly do something else. If you want the default value, simply do not specify `registry`, or use something like `| default(omit)`.

keducoop:
@felixfontein I agree, and I faced that corner case because of our own mistake (setting that variable to an empty value).

But in the other hand our user experience was:

a) Docker login (`docker_login` module) OK
b) Docker run (`docker_container`) KO

Is not expected to get an OK to the login, and then a KO on docker run.

felixfontein:
I agree that this behavior is not great, but I think that this is not the correct fix. The main problem is that Docker daemon does some magic to handle the empty string as a server name (https://github.com/docker/engine/blob/master/registry/service.go#L113) on the login call. This is neither mentioned in the API docs (https://docs.docker.com/engine/api/v1.39/#section/Authentication) nor in the Docker SDK for Python docs (https://docker-py.readthedocs.io/en/stable/api.html#docker.api.daemon.DaemonApiMixin.login).

I think it would be better to emit a warning if that string is passed empty. That would provide feedback to users without making a breaking change.

keducoop:
Looks good for me, thanks for taking the time to point the upstream root cause.

felixfontein:
If you want to create a PR which emits a warning (`module.warn('...')`) when the string is empty, that would be fine for me.

PR #254, opened 2021-12-22
docker_container_exec: allow to specify environment

### LINKED ISSUE #248: Support --env in docker_container_exec
##### SUMMARY
`docker exec` supports an `--env` argument (see https://docs.docker.com/engine/reference/commandline/exec).
Would be nice if the module docker_container_exec would support this too.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
docker_container_exec

felixfontein:
##### SUMMARY
Fixes #248.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
docker_container_exec

### DIFF
--- a/plugins/modules/docker_container_exec.py
+++ b/plugins/modules/docker_container_exec.py
@@ -215,7 +215,7 @@ def main():
     env = client.module.params['env']
 
     if env is not None:
-        for name, value in sorted(env.items()):
+        for name, value in list(env.items()):
             if not isinstance(value, string_types):
                 client.module.fail_json(
                     msg="Non-string value found for env option. Ambiguous env options must be "
--- a/plugins/modules/docker_container_exec.py
+++ b/plugins/modules/docker_container_exec.py
@@ -72,6 +72,14 @@ options:
     default: false
     description:
       - Whether to allocate a TTY.
+  env:
+    description:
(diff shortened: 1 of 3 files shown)


felixfontein:
(Right now includes #253, will rebase once that's merged.)

plugins/modules/docker_container_exec.py
```diff
@@ -183,6 +194,15 @@ def main():
     stdin = client.module.params['stdin']
     strip_empty_ends = client.module.params['strip_empty_ends']
     tty = client.module.params['tty']
+    env = client.module.params['env']
+
+    if env is not None:
+        for name, value in sorted(env.items()):
```
briantist, line 200: not there's anything wrong with it, but just wondering why this is sorted?

briantist (approved):
Looks good to me! No blocking questions or concerns.

plugins/modules/docker_container_exec.py
felixfontein: I think I wanted to write `list`, not `sorted`. The result is the same, except that `sorted` is a bit slower... Will change it.

briantist (approved):

felixfontein:
@briantist thanks for also reviewing this one!

PR #258, opened 2021-12-22
Prepare for distutils.version being removed in Python 3.12

felixfontein:
##### SUMMARY
distutils has been deprecafed and will be removed from Python's stdlib in Python 3.12 (see https://www.python.org/dev/peps/pep-0632/). This PR replaces the use of distutils.version.LooseVersion and distutils.version.StrictVersion with LooseVersion from the vendored copy of distutils.version included with ansible-core 2.12 (https://github.com/ansible/ansible/pull/74644) if available, and falls back to distutils.version for ansible-core 2.11 and before. Since ansible-core 2.11 and earlier do not support Python 3.12 (since they use LooseVersion itself in various places), this incomplete fix should be OK for now. Also, the way this PR works (by adding a new module_utils version that abstracts away where LooseVersion comes from), it is easy to also fix this for ansible-core 2.11 and earlier later on.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Various modules and plugins


briantist (approved):

changelogs/fragments/258-distutils.version.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - "Various modules and plugins - use vendored version of ``distutils.version`` included in ansible-core 2.12 if available. This avoids breakage when ``distutils`` is removed from the standard library of Python 2.12. Note that ansible-core 2.11, ansible-base 2.10 and Ansible 2.9 are right now not compatible with Python 2.12, whence this fix does not target these ansible-core/-base/2.9 versions (https://github.com/ansible-collections/community.docker/pull/258)."
```
felixfontein, line 2:
```suggestion
  - "Various modules and plugins - use vendored version of ``distutils.version`` included in ansible-core 2.12 if available. This avoids breakage when ``distutils`` is removed from the standard library of Python 3.12. Note that ansible-core 2.11, ansible-base 2.10 and Ansible 2.9 are right now not compatible with Python 3.12, whence this fix does not target these ansible-core/-base/2.9 versions (https://github.com/ansible-collections/community.docker/pull/258)."
```

plugins/module_utils/version.py
```diff
@@ -0,0 +1,19 @@
...
+
+from ansible.module_utils.six import raise_from
+
+try:
+    from ansible.module_utils.compat.version import LooseVersion
+except ImportError:
+    try:
+        from distutils.version import LooseVersion
+    except ImportError as exc:
+        raise_from(ImportError('To use this plugin or module with ansible-core < 2.11, you need to use Python < 2.12 with distutils.version present'), exc)
```
felixfontein, line 19:
```suggestion
        raise_from(ImportError('To use this plugin or module with ansible-core < 2.11, you need to use Python < 3.12 with distutils.version present'), exc)
```

felixfontein:

PR #265, opened 2022-01-04
Added cap_add to the documentation for capabilities for easier search

RichardJActon:
##### SUMMARY

Added `cap_add` to the documentation for `capabilities` to make it easier to find the capabilities option if searching based on the docker name for this parameter.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
docker container

### DIFF
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -38,7 +38,7 @@ options:
     type: int
   capabilities:
     description:
-      - List of capabilities to add to the container (equivalent to docker cap_add).
+      - List of capabilities to add to the container.
       - This is equivalent to C(docker run --cap-add), or the docker-compose option C(cap_add).
     type: list
     elements: str
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -38,7 +38,7 @@ options:
     type: int
   capabilities:
     description:
-      - List of capabilities to add to the container.
+      - List of capabilities to add to the container (equivalent to docker cap_add).
     type: list
     elements: str
   cap_drop:
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -39,6 +39,7 @@ options:
   capabilities:
     description:
       - List of capabilities to add to the container (equivalent to docker cap_add).
+      - This is equivalent to C(docker run --cap-add), or the docker-compose option C(cap_add).
     type: list
     elements: str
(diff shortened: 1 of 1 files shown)


plugins/modules/docker_container.py
```diff
@@ -38,7 +38,7 @@
     type: int
   capabilities:
     description:
-      - List of capabilities to add to the container.
+      - List of capabilities to add to the container (equivalent to docker cap_add).
```
felixfontein, line 41: What exactly do you mean with `docker cap_add`? That's apparently not a CLI command (also the CLI option is called `--cap-add`).

RichardJActon: Ah yes looks like I'm conflating the docker-cli & docker-compose syntax. It's `cap_add` in compose for example as used here: https://hub.docker.com/r/pihole/pihole.

Would this be preferable: 

(equivalent to `docker --cap-add`, or `docker-compose` option `cap_add`).

felixfontein: I would mention `docker run --cap-add`, something like this:
```suggestion
      - List of capabilities to add to the container.
      - This is equivalent to C(docker run --cap-add).
```

RichardJActon: This is fine except that the original reason I discovered that this parameter was named differently from docker cli or compose was that when I searched the docs for the literal string `cap_add` whilst trying to translate a docker-compose file to an ansible playbook I didn't get a hit. I suspect others trying to switch to ansible might take the same path so it seems like a good idea to me to have the names of both the cli and compose equivalents in the docs. I'm new here so I don't know if the community has existing conventions around this and having both might get a bit bloated but I'd make it something that has the literal strings from both cli and compose e.g. 

`- This is equivalent to C(docker run --cap-add), or the C(docker-compose) option C(cap_add).`

Happy to go with your suggestion if you think this is still to vague and making it more specific about the compose equivalent would be to verbose.

felixfontein: How about: `- This is equivalent to C(docker run --cap-add), or the docker-compose option C(cap_add).`?

felixfontein: You still need to remove the part from this line :)
```suggestion
      - List of capabilities to add to the container.
```

felixfontein (approved):

felixfontein:
@RichardJActon thanks a lot for improving the docs!

PR #277, opened 2022-01-09
docs build - update references to new repo

briantist:
##### SUMMARY

Related:
- #263 
- https://github.com/ansible-community/community-topics/issues/59

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

docs build workflow

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/.github/workflows/docs-pr.yml
+++ b/.github/workflows/docs-pr.yml
@@ -11,7 +11,7 @@ jobs:
     permissions:
       contents: read
     name: Build Ansible Docs
-    uses: ansible-collections/community.hashi_vault/.github/workflows/_shared-docs-build-pr.yml@main
+    uses: ansible-community/github-docs-build/.github/workflows/_shared-docs-build-pr.yml@main
 
   comment:
     permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-pr.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     permissions:
       contents: read
     name: Build Ansible Docs
-    uses: ansible-collections/community.hashi_vault/.github/workflows/_shared-docs-build-pr.yml@main
+    uses: ansible-community/github-docs-build/.github/workflows/_shared-docs-build-pr.yml@main
```
felixfontein, line 14: I guess at some point it makes sense to pin the version to a tag, but right now while things are still in flux let's better keep `main` :)

felixfontein (approved):

felixfontein:
@briantist thanks!

.github/workflows/docs-pr.yml
felixfontein: Looks like that's not really trivial: https://github.com/ansible-community/github-docs-build/issues/4

briantist: yes, I think while we're getting things set up and ironing out early bugs and such, `main` is better. I consider this in "preview" basically, and anyone who'd like to use it can do so, and if it causes CI failures the entire workflow can be temporarily disabled/enabled in each repo that uses it (in the GitHub UI, commits are not needed). We'll come up with a release strategy at some point; https://github.com/ansible-community/github-docs-build/issues/4 does complicate the matter a bit, it's true.

Issue #246, opened 2021-11-27
docker_secret fails to remove secrets if a service is using them


kaysond:
##### SUMMARY

If you try to change or remove a secret that is being used by an existing service, the module will fail:

`Error removing secret secret_name: 400 Client Error for http+docker://localhost/v1.41/secrets/vurz9eq24g7kyilccxu20m3j3: Bad Request (\"rpc error: code = InvalidArgument desc = secret 'secret_name' is in use by the following service: authelia_authelia\")"}`

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

community.docker.docker_secret

##### ANSIBLE VERSION

```paste below
ansible [core 2.11.1]
  config file = None
  configured module search path = ['/home/administrator/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/administrator/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/administrator/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.8.10 (default, Sep 28 2021, 16:10:42) [GCC 9.3.0]
  jinja version = 2.11.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
# /home/administrator/.ansible/collections/ansible_collections
Collection       Version
---------------- -------
community.docker 2.0.0

# /home/administrator/.local/lib/python3.8/site-packages/ansible_collections
Collection       Version
---------------- -------
community.docker 1.5.0

# /usr/local/lib/python3.8/dist-packages/ansible_collections
Collection       Version
---------------- -------
community.docker 1.9.1
```

##### CONFIGURATION

```paste below
<empty>
```

##### OS / ENVIRONMENT

Ubuntu 20.04, Debian 10

##### STEPS TO REPRODUCE

* Create a secret (probably using ansible so it has the metadata label)
* Create a service that uses that secret (doesnt need to be ansible)
* Change the value of the secret in the ansible module arg, and re-run ansible

##### EXPECTED RESULTS

module successfully changes the secret

##### ACTUAL RESULTS

module fails with the above error

kaysond:
The module itself isn't "smart" (or swarm-aware). It just calls docker-py's [`remove_secret`](https://github.com/docker/docker-py/blob/a48a5a9647761406d66e8271f19fab7fa0c5f582/docker/api/secret.py#L66-L82) method which is the equivalent of running `docker secret rm`

[...]

PR #293, opened 2022-02-05
Add support for rotating docker secrets

### LINKED ISSUE #21: docker_secret.py: use secret versioning to avoid secret deletion on update
##### SUMMARY
The current functionality of docker_secret.py is that when an update to a secret is discovered the existing secret is deleted and the new one is created. This is not ideal since current best practice is to preserve all secret values and increment a version or some other similar metadata to update to a following value. Docker secrets are immutable so deleting them and replacing them breaks their guarantee for rollback scenarios.

I'd like to propose that we retain the existing functionality of delete/replace of updated secrets to not break existing playbooks. Then, add a new module level option, say, `version: true/false` which expects to find a `version: "1"` key/val pair in the `labels` dictionary. Then if you have a secret with `name: foo` and `labels: version: "1"` the module will create an underlying secret with the name of `foo-v1`. Once you change `version: "2"` then you will get a `foo-v2` and `foo-v1` will still be there that can be used to ensure previous configurations work.

It may be that we need to expose a `FOO_VERSION` variable of some sort to tie this into the `docker-compose.yml` file where we have a `name: foo-${FOO_VERSION}` key/val on the `secrets` definition?

I haven't look deeply at the code but I hope this can be done without breaking changes.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
docker_secret

##### ADDITIONAL INFORMATION

Here is an example task.

```yaml
- name: Change the secret data
  docker_secret:
    name: foo
    data: bar
    version: true
    labels:
      version: "1"
    state: present
```

### LINKED ISSUE #109: Support config rotate for docker_config
##### SUMMARY

I would like to be able to change a configuration of a docker service/stack that I deploy via ansible. 

The functionality is described here https://docs.docker.com/engine/swarm/configs/#example-rotate-a-config . 

Right now the docker_config fails to update since a service is using it.
Docker supports create + rm, not update.
Hence, 

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

docker_config

##### ADDITIONAL INFORMATION

Using a playbook like bellow, if I update the config content and run the playbook, it will fail because it tries to update the config (by removing and adding it again).
Removing fails because the configuration is used by the service.

```yaml

- name: Configuration for kibana
  community.docker.docker_config:
    name: kibana_config
    state: present
    data: "{{ lookup('file', 'files/monitoring_elk/kibana.yml') }}"

- name: monitoring_elk
  community.docker.docker_stack:
    state: present
    name: monitoring_elk
    prune: yes
    compose:
      - version: "3.8"
        services:
          elasticsearch:
          kibana:
            image: "docker.elastic.co/kibana/kibana:7.12.0"
            #REDACTED
            configs:
              - source: kibana_config
                target: /usr/share/kibana/config/kibana.yml

```

andrasmaroy:
##### SUMMARY
Add support for rolling updates for docker secrets. This change does roughly what was described in #21.

Adding two additional parameters for the `docker_secret` module:
* `rolling_versions` (default: false): whether to use versioning secrets for rolling updates.
* `versions_to_keep` (default: 5): when using rolling updates, how many of the old versions to keep.

When `rolling_versions` is set to `true` secrets are created with a `_vX` postfix in their names where `X` is replaced with an incremental counter. That counter is also added to the secrets as a label with the name of `ansible_version`. This way when a secret is changed a new version is created with an incremented version number instead of deleting it first then creating again. This aims to solve the problem where the secret is attached to a service, thus deleting it fails.

This is the recommended approach of updating secrets based on the official Docker [documentation](https://docs.docker.com/engine/swarm/secrets/#example-rotate-a-secret).

The change is not breaking, users of this module shouldn't expect any changes.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

docker_secret

##### ADDITIONAL INFORMATION

The problem this PR aims to solve is the following:
When a secret is attached to a service docker won't let it get deleted, thus the current version of the module fails execution if one tries to update a secret that is attached to a service already. The [recommendation from docker](https://docs.docker.com/engine/swarm/secrets/#example-rotate-a-secret) for this is to create a new secret with the updated data then update the service definition adding the new secret in place of the old one. Afterwards the old secret can be safely deleted. Having different names for the secrets is a non-issue since one can define the mountpoint for the secret in the container regardless of its name.

If this change gets accepted I'd implement the same functionality for docker_config as well, as the same issue stands there to fix issue #109.

The PR addresses #21, #246

Fixes #21.

### DIFF
--- a/changelogs/fragments/270-rolling-secrets.yml
+++ b/changelogs/fragments/270-rolling-secrets.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - docker_secret - Add support for rolling update, set ``rolling_versions`` to ``true`` to enable
+  - docker_secret - add support for rolling update, set ``rolling_versions`` to ``true`` to enable (https://github.com/ansible-collections/community.docker/pull/293, https://github.com/ansible-collections/community.docker/issues/21).
--- a/changelogs/fragments/270-rolling-secrets.yml
+++ b/changelogs/fragments/270-rolling-secrets.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - docker_secret - Add support for rolling update, set ``rolling_versions`` to ``true`` to enable
--- a/plugins/modules/docker_secret.py
+++ b/plugins/modules/docker_secret.py
@@ -229,11 +229,19 @@ class SecretManager(DockerBaseClass):
         if self.state == 'present':
             self.data_key = hashlib.sha224(self.data).hexdigest()
             self.present()
+            self.remove_old_versions()
         elif self.state == 'absent':
             self.absent()
 
     def get_version(self, secret):
         return secret.get('Spec', {}).get('Labels', {}).get('version', 0)
+
+    def remove_old_versions(self):
+        if self.versions_to_keep < 0:
+            return
+        if not self.check_mode:
+            while len(self.secrets) >= max(self.versions_to_keep, 1):
+                self.remove_secret(self.secrets.pop(0))
 
     def get_secret(self):
         ''' Find an existing secret. '''
@@ -297,7 +305,8 @@ class SecretManager(DockerBaseClass):
             labels_changed = not compare_generic(self.labels, attrs.get('Labels'), 'allow_more_present', 'dict')
             if data_changed or labels_changed or self.force:
                 # if something changed or force, delete and re-create the secret
-                self.absent()
+                if not self.rolling_versions:
+                    self.absent()
                 secret_id = self.create_secret()
                 self.results['changed'] = True
                 self.results['secret_id'] = secret_id
--- a/plugins/modules/docker_secret.py
+++ b/plugins/modules/docker_secret.py
@@ -352,7 +352,7 @@ def main():
         labels=dict(type='dict'),
         force=dict(type='bool', default=False),
         rolling_versions=dict(type='bool', default=False),
-        versions_to_keep=dict(type='int', default=5)
+        versions_to_keep=dict(type='int', default=5),
     )
 
     required_if = [
--- a/plugins/modules/docker_secret.py
+++ b/plugins/modules/docker_secret.py
@@ -282,7 +282,7 @@ class SecretManager(DockerBaseClass):
         if self.rolling_versions:
             self.version += 1
             labels['ansible_version'] = str(self.version)
-            self.name = self.name + '_v' + str(self.version)
+            self.name = '{name}_v{version}'.format(name=self.name, version=self.version)
         if self.labels:
             labels.update(self.labels)
 
--- a/plugins/modules/docker_secret.py
+++ b/plugins/modules/docker_secret.py
@@ -167,6 +167,12 @@ secret_id:
   returned: success and I(state) is C(present)
   type: str
   sample: 'hzehrmyjigmcp2gb6nlhmjqcv'
+secret_name:
+  description:
+    - The name of the created secret object.
+  returned: success and I(state) is C(present)
+  type: str
+  sample: 'awesome_secret'
 '''
 
 import base64
@@ -298,6 +304,7 @@ class SecretManager(DockerBaseClass):
         ''' Handles state == 'present', creating or updating the secret '''
         if self.secrets:
             self.results['secret_id'] = self.secrets[-1]['ID']
+            self.results['secret_name'] = self.secrets[-1]['Spec']['Name']
             data_changed = False
             attrs = self.secrets[-1].get('Spec', {})
             if attrs.get('Labels', {}).get('ansible_key'):
@@ -316,9 +323,11 @@ class SecretManager(DockerBaseClass):
                 secret_id = self.create_secret()
                 self.results['changed'] = True
                 self.results['secret_id'] = secret_id
+                self.results['secret_name'] = self.name
         else:
             self.results['changed'] = True
             self.results['secret_id'] = self.create_secret()
+            self.results['secret_name'] = self.name
 
     def absent(self):
         ''' Handles state == 'absent', removing the secret '''
@@ -361,7 +370,8 @@ def main():
     try:
         results = dict(
             changed=False,
-            secret_id=''
+            secret_id='',
+            secret_name=''
         )
 
         SecretManager(client, results)()
--- a/plugins/modules/docker_secret.py
+++ b/plugins/modules/docker_secret.py
@@ -54,15 +54,18 @@ options:
   rolling_versions:
     description:
       - If set to C(true), secrets are created with an increasing version number appended to their name.
+      - Adds a label containing the version number to the managed secrets with the name C(ansible_version).
     type: bool
-    default: no
+    default: false
+    version_added: 2.2.0
   versions_to_keep:
     description:
       - When using I(rolling_versions), the number of old versions of the secret to keep.
       - Extraneous old secrets are deleted after the new one is created.
-      - Set to -1 to keep everything or to 0 or 1 to keep only the current one.
+      - Set to C(-1) to keep everything or to C(0) or C(1) to keep only the current one.
     type: int
     default: 5
+    version_added: 2.2.0
   name:
     description:
       - The name of the secret.
@@ -173,6 +176,7 @@ secret_name:
   returned: success and I(state) is C(present)
   type: str
   sample: 'awesome_secret'
+  version_added: 2.2.0
 '''
 
 import base64
--- a/plugins/modules/docker_secret.py
+++ b/plugins/modules/docker_secret.py
@@ -307,10 +307,11 @@ class SecretManager(DockerBaseClass):
     def present(self):
         ''' Handles state == 'present', creating or updating the secret '''
         if self.secrets:
-            self.results['secret_id'] = self.secrets[-1]['ID']
-            self.results['secret_name'] = self.secrets[-1]['Spec']['Name']
+            secret = self.secrets[-1]
+            self.results['secret_id'] = secret['ID']
+            self.results['secret_name'] = secret['Spec']['Name']
             data_changed = False
-            attrs = self.secrets[-1].get('Spec', {})
+            attrs = secret.get('Spec', {})
             if attrs.get('Labels', {}).get('ansible_key'):
                 if attrs['Labels']['ansible_key'] != self.data_key:
                     data_changed = True
@@ -319,7 +320,7 @@ class SecretManager(DockerBaseClass):
                     self.client.module.warn("'ansible_key' label not found. Secret will not be changed unless the force parameter is set to 'yes'")
             labels_changed = not compare_generic(self.labels, attrs.get('Labels'), 'allow_more_present', 'dict')
             if self.rolling_versions:
-                self.version = int(attrs.get('Labels', {}).get('ansible_version', 0))
+                self.version = self.get_version(secret)
             if data_changed or labels_changed or self.force:
                 # if something changed or force, delete and re-create the secret
                 if not self.rolling_versions:
--- a/plugins/modules/docker_secret.py
+++ b/plugins/modules/docker_secret.py
@@ -244,7 +244,7 @@ class SecretManager(DockerBaseClass):
             self.absent()
 
     def get_version(self, secret):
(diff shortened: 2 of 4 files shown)


plugins/modules/docker_secret.py
```diff
@@ -246,39 +293,47 @@ def create_secret(self):
 
         return secret_id
 
+    def remove_secret(self, secret):
+        try:
+            if not self.check_mode:
+                self.client.remove_secret(secret['ID'])
+        except APIError as exc:
+            self.client.fail("Error removing secret %s: %s" % (self.name, to_native(exc)))
```
felixfontein, line 301: Shouldn't it be:
```suggestion
            self.client.fail("Error removing secret %s: %s" % (secret['Spec']['Name'], to_native(exc)))
```
?

felixfontein (commented):
Thanks for your contribution! Please also add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

plugins/modules/docker_secret.py
andrasmaroy: Fixed in [8648e89](https://github.com/ansible-collections/community.docker/pull/293/commits/8648e89).

plugins/modules/docker_secret.py
```diff
@@ -290,7 +345,9 @@ def main():
         data_is_b64=dict(type='bool', default=False),
         data_src=dict(type='path'),
         labels=dict(type='dict'),
-        force=dict(type='bool', default=False)
+        force=dict(type='bool', default=False),
+        rolling_versions=dict(type='bool', default=False),
+        versions_to_keep=dict(type='int', default=5)
```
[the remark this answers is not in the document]
andrasmaroy: Fixed in [d37ccf8](https://github.com/ansible-collections/community.docker/pull/293/commits/d37ccf8).

plugins/modules/docker_secret.py
```diff
@@ -246,39 +293,47 @@ def create_secret(self):
...
+                self.client.remove_secret(secret['ID'])
+        except APIError as exc:
+            self.client.fail("Error removing secret %s: %s" % (self.name, to_native(exc)))
+
     def present(self):
         ''' Handles state == 'present', creating or updating the secret '''
-        secret = self.get_secret()
-        if secret:
-            self.results['secret_id'] = secret['ID']
+        if self.secrets:
```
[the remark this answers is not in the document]
andrasmaroy: Fixed in [0bbf93c](https://github.com/ansible-collections/community.docker/pull/293/commits/0bbf93c).

plugins/modules/docker_secret.py
```diff
@@ -246,39 +293,47 @@ def create_secret(self):
...
+            attrs = self.secrets[-1].get('Spec', {})
             if attrs.get('Labels', {}).get('ansible_key'):
                 if attrs['Labels']['ansible_key'] != self.data_key:
                     data_changed = True
             else:
                 if not self.force:
                     self.client.module.warn("'ansible_key' label not found. Secret will not be changed unless the force parameter is set to 'yes'")
             labels_changed = not compare_generic(self.labels, attrs.get('Labels'), 'allow_more_present', 'dict')
+            if self.rolling_versions:
+                self.version = int(attrs.get('Labels', {}).get('ansible_version', 0))
```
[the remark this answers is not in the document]
andrasmaroy: Fixed in [0bbf93c](https://github.com/ansible-collections/community.docker/pull/293/commits/0bbf93c).

plugins/modules/docker_secret.py
```diff
@@ -232,6 +275,10 @@ def create_secret(self):
         labels = {
             'ansible_key': self.data_key
         }
+        if self.rolling_versions:
+            self.version += 1
+            labels['ansible_version'] = str(self.version)
+            self.name = self.name + '_v' + str(self.version)
```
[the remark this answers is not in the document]
andrasmaroy: Fixed in [c90ded3](https://github.com/ansible-collections/community.docker/pull/293/commits/c90ded3).

plugins/modules/docker_secret.py
```diff
@@ -204,26 +222,51 @@ def __init__(self, client, results):
...
+        self.get_secret()
         if self.state == 'present':
             self.data_key = hashlib.sha224(self.data).hexdigest()
             self.present()
+            self.remove_old_versions()
         elif self.state == 'absent':
             self.absent()
 
+    def get_version(self, secret):
+        return secret.get('Spec', {}).get('Labels', {}).get('version', 0)
```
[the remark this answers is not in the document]
andrasmaroy: Fixed in [0e0429c](https://github.com/ansible-collections/community.docker/pull/293/commits/0e0429c).

plugins/modules/docker_secret.py
```diff
@@ -155,6 +167,12 @@
   returned: success and I(state) is C(present)
   type: str
   sample: 'hzehrmyjigmcp2gb6nlhmjqcv'
+secret_name:
+  description:
+    - The name of the created secret object.
+  returned: success and I(state) is C(present)
+  type: str
+  sample: 'awesome_secret'
```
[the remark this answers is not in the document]
andrasmaroy: Fixed in [5c2a3ed](https://github.com/ansible-collections/community.docker/pull/293/commits/5c2a3ed).

plugins/modules/docker_secret.py
```diff
@@ -51,6 +51,18 @@
       - If C(true), an existing secret will be replaced, even if it has not changed.
     type: bool
     default: no
+  rolling_versions:
+    description:
+      - If set to C(true), secrets are created with an increasing version number appended to their name.
```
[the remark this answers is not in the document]
andrasmaroy: Fixed in [5c2a3ed](https://github.com/ansible-collections/community.docker/pull/293/commits/5c2a3ed).

plugins/modules/docker_secret.py
```diff
@@ -51,6 +51,18 @@
...
+    type: bool
+    default: no
+  versions_to_keep:
+    description:
+      - When using I(rolling_versions), the number of old versions of the secret to keep.
+      - Extraneous old secrets are deleted after the new one is created.
+      - Set to -1 to keep everything or to 0 or 1 to keep only the current one.
```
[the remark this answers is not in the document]
andrasmaroy: Fixed in [5c2a3ed](https://github.com/ansible-collections/community.docker/pull/293/commits/5c2a3ed).

plugins/modules/docker_secret.py
```diff
@@ -51,6 +51,18 @@
...
+    type: int
+    default: 5
```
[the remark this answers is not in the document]
andrasmaroy: Fixed in [5c2a3ed](https://github.com/ansible-collections/community.docker/pull/293/commits/5c2a3ed).

plugins/modules/docker_secret.py
```diff
@@ -51,6 +51,18 @@
       - If C(true), an existing secret will be replaced, even if it has not changed.
     type: bool
     default: no
+  rolling_versions:
+    description:
+      - If set to C(true), secrets are created with an increasing version number appended to their name.
+    type: bool
+    default: no
```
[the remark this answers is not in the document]
andrasmaroy: Fixed in [5c2a3ed](https://github.com/ansible-collections/community.docker/pull/293/commits/5c2a3ed).

plugins/modules/docker_secret.py
```diff
@@ -204,26 +226,51 @@ def __init__(self, client, results):
...
+        self.get_secret()
         if self.state == 'present':
             self.data_key = hashlib.sha224(self.data).hexdigest()
             self.present()
+            self.remove_old_versions()
         elif self.state == 'absent':
             self.absent()
 
+    def get_version(self, secret):
+        return int(secret.get('Spec', {}).get('Labels', {}).get('ansible_version', 0))
```
felixfontein, line 247:
```suggestion
        try:
            return int(secret.get('Spec', {}).get('Labels', {}).get('ansible_version', 0))
        except ValueError:
            return 0
```
In case someone adds their own `ansible_version` label with something that's not an integer :)

plugins/modules/docker_secret.py
```diff
@@ -204,26 +226,51 @@ def __init__(self, client, results):
...
+
+    def remove_old_versions(self):
+        if self.versions_to_keep < 0:
```
felixfontein, line 250:
```suggestion
        if not rolling_versions or self.versions_to_keep < 0:
```

plugins/modules/docker_secret.py
```diff
@@ -204,26 +226,51 @@ def __init__(self, client, results):
...
+            return
+        if not self.check_mode:
+            while len(self.secrets) >= max(self.versions_to_keep, 1):
```
felixfontein, line 253:
```suggestion
            while len(self.secrets) > max(self.versions_to_keep, 1):
```
You only want to remove something if there are too many secrets.

plugins/modules/docker_secret.py
```diff
@@ -232,6 +279,10 @@ def create_secret(self):
         labels = {
             'ansible_key': self.data_key
         }
+        if self.rolling_versions:
+            self.version += 1
+            labels['ansible_version'] = str(self.version)
+            self.name = '{name}_v{version}'.format(name=self.name, version=self.version)
```
felixfontein, line 285: For the above to work correctly, you need to add something here though.
```suggestion
            self.name = '{name}_v{version}'.format(name=self.name, version=self.version)
            self.secrets.append(...)
```

plugins/modules/docker_secret.py
andrasmaroy: The newly created secret is not part of this list, hence `>=`.

plugins/modules/docker_secret.py
andrasmaroy: Yeah, I was thinking about doing this, but to get the correct structure we'd have to run a `client.get_secret()` which seemed to me to be more complex than what's necessary.

plugins/modules/docker_secret.py
felixfontein: But if no new secret is created due to idempotency, one too many will be removed here.

plugins/modules/docker_secret.py
andrasmaroy: Indeed, fixed in [c0ccb48](https://github.com/ansible-collections/community.docker/pull/293/commits/c0ccb48).

plugins/modules/docker_secret.py
andrasmaroy: Good catch, fixed in [c05337b](https://github.com/ansible-collections/community.docker/pull/293/commits/c05337b).

plugins/modules/docker_secret.py
andrasmaroy: That's correct, let me look into fixing this.

andrasmaroy: Okay, fixed in [da53c00](https://github.com/ansible-collections/community.docker/pull/293/commits/da53c00) seems to be working correctly with quick testing.

felixfontein:
I think it looks good now. The only part missing is a changelog fragment.

changelogs/fragments/270-rolling-secrets.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - docker_secret - Add support for rolling update, set ``rolling_versions`` to ``true`` to enable
```
felixfontein, line 2:
```suggestion
  - docker_secret - add support for rolling update, set ``rolling_versions`` to ``true`` to enable (https://github.com/ansible-collections/community.docker/pull/293, https://github.com/ansible-collections/community.docker/issues/21).
```

felixfontein (approved):
I think it looks good. I will merge if nobody complains until, say, next Saturday.

WojciechowskiPiotr:
shipit :) I'm not merging it to let others review it as well :)

felixfontein:
@andrasmaroy thanks for implementing this!

PR #294, opened 2022-02-05
Implement `cap_add` and `cap_drop` handling for `docker_swarm_service`

andrasmaroy:
##### SUMMARY
With the [5.0.3 release of Docker SDK](https://github.com/docker/docker-py/releases/tag/5.0.3) for Python support was added for these parameters for swarm services. With that adding support in Ansible should be as simple as exposing these variables for the module. That's what's done in this change.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

docker_swarm_service

### DIFF
--- a/plugins/modules/docker_swarm_service.py
+++ b/plugins/modules/docker_swarm_service.py
@@ -669,6 +669,10 @@ swarm_service:
     "args": [
       "3600"
     ],
+    "cap_add": null,
+    "cap_drop": [
+      "ALL"
+    ],
     "command": [
       "sleep"
     ],
--- a/plugins/modules/docker_swarm_service.py
+++ b/plugins/modules/docker_swarm_service.py
@@ -627,6 +627,18 @@ options:
       - Path to the working directory.
       - Corresponds to the C(--workdir) option of C(docker service create).
     type: str
+  cap_add:
+    description:
+      - List of capabilities to add to the container.
+      - Requires API version >= 1.41.
+    type: list
+    elements: str
+  cap_drop:
+    description:
+      - List of capabilities to drop from the container.
+      - Requires API version >= 1.41.
+    type: list
+    elements: str
 extends_documentation_fragment:
 - community.docker.docker
 - community.docker.docker.docker_py_2_documentation
@@ -1240,6 +1252,8 @@ class DockerService(DockerBaseClass):
         self.update_order = None
         self.working_dir = None
         self.init = None
+        self.cap_add = None
+        self.cap_drop = None
 
         self.docker_api_version = docker_api_version
         self.docker_py_version = docker_py_version
@@ -1296,6 +1310,8 @@ class DockerService(DockerBaseClass):
             'user': self.user,
             'working_dir': self.working_dir,
             'init': self.init,
+            'cap_add': self.cap_add,
+            'cap_drop': self.cap_drop,
         }
 
     @property
@@ -1525,6 +1541,8 @@ class DockerService(DockerBaseClass):
         s.working_dir = ap['working_dir']
         s.read_only = ap['read_only']
         s.init = ap['init']
+        s.cap_add = ap['cap_add']
+        s.cap_drop = ap['cap_drop']
 
         s.networks = get_docker_networks(ap['networks'], network_ids)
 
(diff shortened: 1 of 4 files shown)


tests/integration/targets/docker_swarm_service/tasks/tests/options.yml
```diff
@@ -1920,3 +1920,79 @@
...
+    state: absent
+  diff: no
+
+- assert:
+    that:
+    - capabilities_1 is changed
+    - capabilities_2 is not changed
+    - capabilities_3 is not changed
+    - capabilities_4 is changed
+  when: docker_py_version is version('5.0.3', '>=')
```
felixfontein, line 1991: You need to check the API version here as well, in the CI failures it has a new enough docker-py, but not a new enough API version.

andrasmaroy:
I'm a bit confused about these CI failures, the errors seem unrelated to me, am I just not seeing the full picture?

felixfontein:
The CI errors are caused by your changes, see my explanation here: https://github.com/ansible-collections/community.docker/pull/294#discussion_r800148713 Capabilities for swarm services need both a new enough Docker SDK for Python version **and** a new enough API version. The former is available on the CI nodes which fail, the latter is not.

tests/integration/targets/docker_swarm_service/tasks/tests/options.yml
```diff
@@ -1920,3 +1920,79 @@
...
+    - capabilities_2 is not changed
+    - capabilities_3 is not changed
+    - capabilities_4 is changed
+  when: docker_api_version is version('1.41', '>=') and docker_py_version is version('5.0.3', '>=')
+
+- assert:
+    that:
+    - capabilities_1 is failed
+    - "('version is ' ~ docker_py_version ~ ' ') in capabilities_1.msg"
+    - "'Minimum version required is 5.0.3 ' in capabilities_1.msg"
```
felixfontein, line 1997: The error message that is failing CI is `Docker API version is 1.40. Minimum version required is 1.41 to set cap_add option.`:
```suggestion
    - >
      (('version is ' ~ docker_py_version ~ ' ') in capabilities_1.msg and 'Minimum version required is 5.0.3 ' in capabilities_1.msg)
      or (('Docker API version is ' ~ docker_api_version ~ '. ') in capabilities_1.msg and 'Minimum version required is 1.41 ' in capabilities_1.msg)
```

felixfontein:
BTW, you also need to mention the two new return values in the `swarm_service` return value example.

tests/integration/targets/docker_swarm_service/tasks/tests/options.yml
andrasmaroy: Shoot I missed this, thanks for the explanation!

plugins/modules/docker_swarm_service.py
```diff
@@ -1773,6 +1797,10 @@ def compare(self, os):
             force_update = True
         if self.init is not None and self.init != os.init:
             differences.add('init', parameter=self.init, active=os.init)
+        if self.cap_add is not None and self.cap_add != os.cap_add:
```
felixfontein, line 1800: Lists should be compared with `has_list_changed`. It also takes care of sorting (I assume the order of the capabilities does not matter; if it does, add `sort_lists=False`):
```suggestion
        if has_list_changed(self.cap_add, os.cap_add):
```

plugins/modules/docker_swarm_service.py
```diff
@@ -1773,6 +1797,10 @@ def compare(self, os):
...
+            differences.add('cap_add', parameter=self.cap_add, active=os.cap_add)
+        if self.cap_drop is not None and self.cap_drop != os.cap_drop:
```
felixfontein, line 1802:
```suggestion
        if has_list_changed(self.cap_drop, os.cap_drop):
```

felixfontein (approved):
Looks good to me. I would be glad if someone more familiar with this module could also take a look at this :)

If nobody does, I'll also merge this one in roughly a week (probably next Saturday).

WojciechowskiPiotr:
lgtm

felixfontein:
@andrasmaroy also thanks for this contribution!
@WojciechowskiPiotr thanks again for reviewing!

PR #295, opened 2022-02-05
Add support for rotating docker configs

andrasmaroy:
##### SUMMARY
Add support for rolling updates for docker configs. Based on the same functionality for secrets in #293.

Adding two additional parameters for the `docker_config` module:
* `rolling_versions` (default: false): whether to use versioning configs for rolling updates.
* `versions_to_keep` (default: 5): when using rolling updates, how many of the old versions to keep.

When `rolling_versions` is set to `true` configs are created with a `_vX` postfix in their names where `X` is replaced with an incremental counter. That counter is also added to the configs as a label with the name of `ansible_version`. This way when a config is changed a new version is created with an incremented version number instead of deleting it first then creating again. This aims to solve the problem where the config is attached to a service, thus deleting it fails.

This is the recommended approach of updating configs based on the official Docker [documentation](https://docs.docker.com/engine/swarm/configs/#example-rotate-a-config).

The change is not breaking, users of this module shouldn't expect any changes.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

docker_config

##### ADDITIONAL INFORMATION

The problem this PR aims to solve is the following:
When a config is attached to a service docker won't let it get deleted, thus the current version of the module fails execution if one tries to update a config that is attached to a service already. The [recommendation from docker](https://docs.docker.com/engine/swarm/configs/#example-rotate-a-config) for this is to create a new config with the updated data then update the service definition adding the new config in place of the old one. Afterwards the old config can be safely deleted. Having different names for the config is a non-issue since one can define the mountpoint for the config in the container regardless of its name.

The PR addresses #109

### DIFF
--- a/changelogs/fragments/272-rolling-configs.yml
+++ b/changelogs/fragments/272-rolling-configs.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - docker_config - add support for rolling update, set ``rolling_versions`` to ``true`` to enable (https://github.com/ansible-collections/community.docker/pull/295, https://github.com/ansible-collections/community.docker/issues/109).
--- a/plugins/modules/docker_config.py
+++ b/plugins/modules/docker_config.py
@@ -51,6 +51,21 @@ options:
       - If C(true), an existing config will be replaced, even if it has not been changed.
     type: bool
     default: no
+  rolling_versions:
+    description:
+      - If set to C(true), configs are created with an increasing version number appended to their name.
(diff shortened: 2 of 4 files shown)


felixfontein (approved):
I think it looks good. I will merge if nobody complains until, say, next Saturday.

WojciechowskiPiotr:
shipit

felixfontein:
@andrasmaroy thanks for your contribution!

PR #297, opened 2022-02-10
docker connection plugin: fix config docs and update to use config system

### LINKED ISSUE #307: New warning about connection plugin
##### SUMMARY
Using Ansible core devel branch, I see a warning:

> [WARNING]: The "community.docker.docker" connection plugin has an improperly configured remote target value, forcing "inventory_hostname" templated value instead of the string

This issues proposes to fix whatever this message is advising to fix.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
plugins/connection/docker.py

##### ANSIBLE VERSION
```paste below
$ ansible --version
[WARNING]: You are running the development version of Ansible. You should only run Ansible from "devel" if you are modifying the Ansible engine, or
trying out features under development. This is a rapidly changing source of code and can become unstable at any point.
ansible [core 2.13.0.dev0] (devel 27923aad7e) last updated 2022/03/08 09:03:09 (GMT -400)
  config file = /home/alancoding/repos/playbooks/ansible.cfg
  configured module search path = ['/home/alancoding/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/alancoding/repos/ansible/lib/ansible
  ansible collection location = /home/alancoding/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/alancoding/repos/awx/env/bin/ansible
  python version = 3.10.1 (main, Jan 10 2022, 00:00:00) [GCC 11.2.1 20211203 (Red Hat 11.2.1-7)]
  jinja version = 3.0.3
  libyaml = True
```

##### COLLECTION VERSION
```
$ ansible-galaxy collection list
[WARNING]: You are running the development version of Ansible. You should only run Ansible from "devel" if you are modifying the Ansible engine, or
trying out features under development. This is a rapidly changing source of code and can become unstable at any point.

# /home/alancoding/.ansible/collections/ansible_collections
Collection                      Version
------------------------------- -------
ansible_runner.display_callback 1.0.0  
awx.awx                         19.4.0 
cloud.common                    2.0.3  
community.docker                2.2.1  
community.kubernetes            2.0.0  
containers.podman               1.9.1  
kubernetes.core                 2.0.2  
openstack.cloud                 1.7.0  

# /usr/share/ansible/collections/ansible_collections
Collection Version
---------- -------
awx.awx    19.4.1 
```

##### CONFIGURATION
defaults

##### OS / ENVIRONMENT
Fedora host machine, CentOS stream derived target containers.

##### STEPS TO REPRODUCE
My inventory looks like this:

```
$ ansible-inventory ~/repos/tower-qa/playbooks/inventory-docker.py --list --export
[WARNING]: You are running the development version of Ansible. You should only run Ansible from "devel" if you are modifying the Ansible engine, or
trying out features under development. This is a rapidly changing source of code and can become unstable at any point.
{
    "_meta": {
        "hostvars": {
            "127.0.0.1": {
                "ansible_connection": "local",
                "ansible_python_interpreter": "/home/alancoding/repos/awx/env/bin/python3"
            },
            "tools_awx_1": {
                "ansible_connection": "community.docker.docker",
                "ansible_user": "0"
            }
        }
    },
    "all": {
        "children": [
            "automationcontroller",
            "execution_nodes",
            "local",
            "sso",
            "tower",
            "ungrouped"
        ]
    },
    "automationcontroller": {
        "hosts": [
            "tools_awx_1"
        ]
    },
    "local": {
        "hosts": [
            "127.0.0.1"
        ]
    },
    "tower": {
        "hosts": [
            "tools_awx_1"
        ]
    }
}
```

No warnings are raised parsing the inventory.

Let's run a playbook that does work against those containers. I'm running https://github.com/AlanCoding/playbooks/blob/devel/image_distribute.yml

```
ansible-playbook image_distribute.yml --tags list
```

(ultimately, you'll get the same result with any `comand:` task, the problem appears to be in the hostvars as the connection plugin gets used)

##### EXPECTED RESULTS
Works with no warnings

##### ACTUAL RESULTS
Works but has warnings

```
TASK [Get the images on the system, double-check your work] **********************************************************************************************
[WARNING]: The "community.docker.docker" connection plugin has an improperly configured remote target value, forcing "inventory_hostname" templated value
instead of the string
changed: [tools_awx_1]
```

This warning comes from

https://github.com/ansible/ansible/blob/56edbd2bbb372a61dae2017923f1d8e33d1922d9/lib/ansible/playbook/play_context.py#L300-L305

Introduced in https://github.com/ansible/ansible/pull/77005

The Ansible issue https://github.com/ansible/ansible/issues/77001 gives a little bit more description of the problem. Mentioned in that issue:

> I suspect i messed up the default play_context.remote_addr and the docker connection plugin is still using that instead of 'get_option'. That we get the error as module output makes no sense.

It is true, this collection sets this option:

https://github.com/ansible-collections/community.docker/blob/4b235a260737de318fd95453c0146534b199f10f/plugins/connection/docker.py#L35-L41

bcoca:
wean off play_context which did not have the correct data in all cases

Fixes #307.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

connection/docker.py

### DIFF
--- a/plugins/connection/docker.py
+++ b/plugins/connection/docker.py
@@ -123,12 +123,13 @@ class Connection(ConnectionBase):
             except ValueError:
                 raise AnsibleError("docker command not found in PATH")
 
-        docker_version = self._get_docker_version()
-        if docker_version == u'dev':
+        self.docker_version = self._get_docker_version()
+        if self.docker_version == u'dev':
             display.warning(u'Docker version number is "dev". Will assume latest version.')
-        if docker_version != u'dev' and LooseVersion(docker_version) < LooseVersion(u'1.3'):
+        if self.docker_version != u'dev' and LooseVersion(self.docker_version) < LooseVersion(u'1.3'):
             raise AnsibleError('docker connection type requires docker 1.3 or higher')
 
+        # The remote user
         self._set_conn_data()
 
     @staticmethod
@@ -218,19 +219,18 @@ class Connection(ConnectionBase):
         # docker arguments
         self._docker_args = []
         if self.get_option('docker_extra_args'):
-            cmd_args += self.get_option('docker_extra_args').split(' ')
-        elif self._play_context.get('docker_extra_args').
-            cmd_args += self._play_context.get('docker_extra_args').split(' ')
+            self._docker_args += self.get_option('docker_extra_args').split(' ')
+        elif self._play_context.get('docker_extra_args'):
+            self._docker_args += self._play_context.get('docker_extra_args').split(' ')
 
-        # The remote user
         self.remote_user = self.get_option('remote_user')
-        if remote_user is None and self._play_context.remote_user is not None:
+        if self.remote_user is None and self._play_context.remote_user is not None:
             self.remote_user = self._play_context.remote_user
         # The actual user which will execute commands in docker (if known)
         self.actual_user = None
 
         if self.remote_user is not None:
-            if docker_version == u'dev' or LooseVersion(docker_version) >= LooseVersion(u'1.7'):
+            if self.docker_version == u'dev' or LooseVersion(self.docker_version) >= LooseVersion(u'1.7'):
                 # Support for specifying the exec user was added in docker 1.7
                 self.actual_user = self.remote_user
             else:
@@ -238,7 +238,7 @@ class Connection(ConnectionBase):
                 self.actual_user = self._get_docker_remote_user()
                 if self.actual_user != self.get_option('remote_user'):
                     display.warning(u'docker {0} does not support remote_user, using container default: {1}'
-                                    .format(docker_version, self.actual_user or u'?'))
+                                    .format(self.docker_version, self.actual_user or u'?'))
         elif self._display.verbosity > 2:
             # Since we're not setting the actual_user, look it up so we have it for logging later
             # Only do this if display verbosity is high enough that we'll need the value
@@ -257,8 +257,6 @@ class Connection(ConnectionBase):
             display.vvv(u"ESTABLISH DOCKER CONNECTION FOR USER: {0}".format(
                 self.actual_user or u'?'), host=self.get_option('remote_addr')
             )
-
-
             self._connected = True
 
     def exec_command(self, cmd, in_data=None, sudoable=False):
--- a/plugins/connection/docker.py
+++ b/plugins/connection/docker.py
@@ -32,8 +32,8 @@ DOCUMENTATION = '''
             - name: ansible_docker_host
       remote_user:
         description:
-            - The user to execute as inside the container
-            - If docker is not configured to allow this, the one set by docker itself will be used
+            - The user to execute as inside the container.
+            - If docker is not configured to allow this, the one set by docker itself will be used.
         vars:
             - name: ansible_user
             - name: ansible_docker_user
--- a/plugins/connection/docker.py
+++ b/plugins/connection/docker.py
@@ -62,6 +62,7 @@ DOCUMENTATION = '''
         env:
             - name: ANSIBLE_TIMEOUT
             - name: ANSIBLE_DOCKER_TIMEOUT
+              version_added: 2.2.0
         ini:
             - key: timeout
               section: defaults
--- a/plugins/connection/docker.py
+++ b/plugins/connection/docker.py
@@ -48,7 +48,7 @@ DOCUMENTATION = '''
             - name: remote_user
       docker_extra_args:
         description:
-            - Extra arguments to pass to the docker command line
+            - Extra arguments to pass to the docker command line.
         default: ''
         vars:
             - name: ansible_docker_extra_args
--- a/plugins/connection/docker.py
+++ b/plugins/connection/docker.py
@@ -114,6 +114,8 @@ class Connection(ConnectionBase):
         # configured to be connected to by root and they are not running as
         # root.
 
+        self._docker_args = []
+
         # Windows uses Powershell modules
         if getattr(self._shell, "_IS_WINDOWS", False):
             self.module_implementation_preferences = ('.ps1', '.exe', '')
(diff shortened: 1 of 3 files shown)


plugins/connection/docker.py
```diff
@@ -37,8 +37,26 @@
             - The name of the container you want to access.
         default: inventory_hostname
         vars:
+            - name: inventory_hostname
```
fao89, line 40: 👍🏽

fao89 (approved):

plugins/connection/docker.py
```diff
@@ -22,23 +22,56 @@
...
+            - The name of the container you want to access.
+        default: inventory_hostname
+        vars:
+            - name: inventory_hostname
+            - name: ansible_host
+            - name: ansible_docker_host
       remote_user:
         description:
             - The user to execute as inside the container
+            - If docker is not configured to allow this, the one set by docker itself will be used
```
felixfontein, line 36:
```suggestion
            - The user to execute as inside the container.
            - If docker is not configured to allow this, the one set by docker itself will be used.
```

felixfontein:
Hmm, one thing I'm wondering is whether this actually works with **older** versions of ansible-core than 2.13 / devel branch.

My main issue here are when values are specified as CLI args and task keywords. The old code should happily use such values as ansible-core somehow placed them in `_play_context`.

briantist:
Does this change behavior because of https://github.com/ansible/ansible/issues/73268 ? 
As in, does getting the vars from `_play_context` give the result from processing templating, or does it (like current behavior) give you raw values without jinja processing?

bcoca:
@briantist unrelated, you never get vars from play_context .. it does UPDATE the task variables with some 'connection vars' but not always correctly, specially in loop or delegation context (things i think we finally fixed in current devel)

lookups are normally in charge of templating their own vars, but config should be doing it for it in this case, so its a different bug.

plugins/connection/docker.py
```diff
@@ -22,23 +22,56 @@
           R(community.docker.docker_api,ansible_collections.community.docker.docker_api_connection)
           connection plugin.
     options:
+      remote_addr:
+        description:
+            - The name of the container you want to access.
+        default: inventory_hostname
```
felixfontein, line 28: I don't think it makes sense to keep this.
```suggestion
```

tests/unit/plugins/connection/test_docker.py
```diff
@@ -41,6 +41,7 @@ def setUp(self):
     def tearDown(self):
         pass
 
+    @mock.patch('ansible_collections.community.docker.plugins.connection.docker._load_name', 'community.docker.docker')
```
felixfontein, line 44: You might need to do `from ansible_collections.community.docker.plugins.connection import docker as docker_plugin` at the top and then `docker_plugin._load_name = 'community.docker.docker'`. `@mock.patch()` only wants to change properties that already exist.

felixfontein:
(If backported this needs to be done manually, if just to remove the new options.)

plugins/connection/docker.py
bcoca: it does inform about the correct default, even if functionally it cannot ever result in the string value

plugins/connection/docker.py
```diff
@@ -167,7 +181,7 @@ def _get_docker_version(self):
 
     def _get_docker_remote_user(self):
         """ Get the default user configured in the docker container """
-        p = subprocess.Popen([self.docker_cmd, 'inspect', '--format', '{{.Config.User}}', self._play_context.remote_addr],
+        p = subprocess.Popen([self.docker_cmd, 'inspect', '--format', '{{.Config.User}}', self.get_option('remote_addr')],
```
felixfontein, line 184: @bcoca here `self.get_option('remote_addr')` always evaluates to `"inventory_hostname"`, which makes Ansible print `[WARNING]: unable to retrieve default user from docker container:   Error: No such object: inventory_hostname` for every task.

bcoca: the vars entry should ensure you don't get 'invenetory_hostname' but the actual value, but  that requires variables to be available, probably need a pc fallback

AlanCoding:
Tested updated file and it works now, and warning is resolved.

plugins/connection/docker.py
felixfontein: This happened when it was called from `__init__()`, despite it said it should not be called from `__init__()` :) I removed that call.

I also made sure that extra args aren't repeatedly added, and that the user retrieval is cached (so it's not called multiple times per task).

felixfontein (approved):
Looks good to me. @AlanCoding can you try again (just to make sure I didn't screw something up)? @bcoca can you also take a quick look? Thanks!

AlanCoding:
:+1: latest version of the plugin is working just fine as well.

plugins/connection/docker.py
```diff
@@ -115,6 +115,7 @@ def __init__(self, play_context, new_stdin, *args, **kwargs):
         # root.
 
         self._docker_args = []
+        self._container_user_cache = {}
```
bcoca, line 118: do not cache user as it is not guaranteed to be the same for all calls in the task

plugins/connection/docker.py
```diff
@@ -226,7 +226,7 @@ def _set_conn_data(self):
 
         # TODO: this is mostly for backwards compatibility, play_context is used as fallback for older versions
         # docker arguments
-        self._docker_args.clear()
+        del self._docker_args[:]
```
bcoca, line 229: Is this not supposed to work with 2.9?

felixfontein: Do you mean Ansible 2.9? How's this related to Ansible 2.9?

plugins/connection/docker.py
felixfontein: Do you mean that someone could recreate the container during the playbook/role run?

plugins/connection/docker.py
bcoca: nvmd, this is a local array ... had confused with pc 'special' structuring of attirbutes

felixfontein (approved):

felixfontein:
@bcoca thanks a lot for fixing this!
@fao89 @AlanCoding thanks for reviewing/testing!

plugins/connection/docker.py
bcoca: no, uninstall and reinstall the docker tools at the controller

PR #311, opened 2022-03-21
docker connection plugin: simplify actual_user handling code

felixfontein:
##### SUMMARY
`self.actual_user` was only used in exactly once place. Make sure it is only computed then.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
docker connection plugin


briantist (approved):
much cleaner. nice

felixfontein:

PR #312, opened 2022-03-21
Connection plugins: implement connection reset

felixfontein:
##### SUMMARY
The default behavior for connection plugins which do not implement `reset()` is `display.warning("Reset is not implemented for this connection")`. A proper reset isn't really needed, so instead let's do something that makes sense.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
docker and docker_api connection plugins


briantist (approved):

felixfontein:

PR #345, opened 2022-05-03
Support --template-driver for docker_config (#332)

### LINKED ISSUE #332: Support --template-driver for docker_config
##### SUMMARY

The `--template-driver` parameter set to `golang` when creating a docker config means we can create docker configs referencing environment variables, secrets and other information. Support for setting this in the `docker_config` module would be useful.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

docker_config

##### ADDITIONAL INFORMATION

```yaml

```

sashajenner:
##### SUMMARY
Adding `template_driver` option with one option 'golang'. Creates a dictionary `templating` or `None` as required by `create_config` (https://docker-py.readthedocs.io/en/stable/api.html#docker.api.config.ConfigApiMixin.create_config). This is to be used in conjunction with a template file specified in `data`.

Fixes  #332.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Support --template-driver for docker_config

### DIFF
--- a/plugins/modules/docker_config.py
+++ b/plugins/modules/docker_config.py
@@ -81,7 +81,7 @@ options:
       - present
   template_driver:
     description:
-      - TODO
+      - Set to C(golang) to use a Golang template file in I(data).
     type: str
     choices:
       - golang
--- a/plugins/modules/docker_config.py
+++ b/plugins/modules/docker_config.py
@@ -289,8 +289,7 @@ class ConfigManager(DockerBaseClass):
         config_id = None
         # We can't see the data after creation, so adding a label we can use for idempotency check
         labels = {
-            'ansible_key': self.data_key,
-            'template_driver': self.template_driver,
+            'ansible_key': self.data_key
         }
         if self.rolling_versions:
             self.version += 1
@@ -339,7 +338,12 @@ class ConfigManager(DockerBaseClass):
             else:
                 if not self.force:
                     self.client.module.warn("'ansible_key' label not found. Config will not be changed unless the force parameter is set to 'yes'")
-            if attrs['Labels']['template_driver'] != self.template_driver:
+            # template_driver has changed if it was set in the previous config
+            # and now it differs, or if it wasn't set but now it is.
+            if attrs.get('Templating', {}).get('Name'):
+                if attrs['Templating']['Name'] != self.template_driver:
+                    template_driver_changed = True
+            elif self.template_driver:
                 template_driver_changed = True
             labels_changed = not compare_generic(self.labels, attrs.get('Labels'), 'allow_more_present', 'dict')
             if self.rolling_versions:
--- a/plugins/modules/docker_config.py
+++ b/plugins/modules/docker_config.py
@@ -81,10 +81,11 @@ options:
       - present
   template_driver:
     description:
-      - Set to C(golang) to use a Golang template file in I(data).
+      - Set to C(golang) to use a Go template in I(data) or a Go template file in I(data_src).
     type: str
     choices:
       - golang
+    version_added: 2.5.0
 
 extends_documentation_fragment:
 - community.docker.docker
@@ -98,7 +99,6 @@ requirements:
 author:
   - Chris Houseknecht (@chouseknecht)
   - John Hu (@ushuz)
-  - Sasha Jenner (@sashajenner)
 '''
 
 EXAMPLES = '''
@@ -306,8 +306,11 @@ class ConfigManager(DockerBaseClass):
 
         try:
             if not self.check_mode:
-                config_id = self.client.create_config(self.name, self.data,
-                    labels=labels, templating=self.templating)
+                # only use templating argument when self.templating is defined
+                kwargs = {}
+                if self.templating:
+                    kwargs['templating'] = self.templating
+                config_id = self.client.create_config(self.name, self.data, labels=labels, **kwargs)
                 self.configs += self.client.configs(filters={'id': config_id})
         except APIError as exc:
             self.client.fail("Error creating config: %s" % to_native(exc))
@@ -384,6 +387,10 @@ def main():
         ('data', 'data_src'),
     ]
 
+    option_minimal_versions = dict(
+        template_driver=dict(docker_py_version='5.0.3', docker_api_version='1.37'),
+    )
+
     client = AnsibleDockerClient(
         argument_spec=argument_spec,
         supports_check_mode=True,
@@ -391,6 +398,7 @@ def main():
         mutually_exclusive=mutually_exclusive,
         min_docker_version='2.6.0',
         min_docker_api_version='1.30',
+        option_minimal_versions=option_minimal_versions,
     )
 
     try:
--- a/plugins/modules/docker_config.py
+++ b/plugins/modules/docker_config.py
@@ -331,6 +331,7 @@ class ConfigManager(DockerBaseClass):
             self.results['config_id'] = config['ID']
             self.results['config_name'] = config['Spec']['Name']
             data_changed = False
+            template_driver_changed = False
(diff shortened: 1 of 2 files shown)


plugins/modules/docker_config.py
```diff
@@ -79,6 +79,12 @@
     choices:
       - absent
       - present
+  template_driver:
+    description:
+      - Set to C(golang) to use a Golang template file in I(data).
+    type: str
+    choices:
+      - golang
```
felixfontein, line 87:
```suggestion
      - golang
    version_added: 2.5.0
```

plugins/modules/docker_config.py
```diff
@@ -79,6 +79,12 @@
     choices:
       - absent
       - present
+  template_driver:
+    description:
+      - Set to C(golang) to use a Golang template file in I(data).
```
felixfontein, line 84:
```suggestion
      - Set to C(golang) to use a Go template in I(data) or a template file in I(data_src).
```

plugins/modules/docker_config.py
```diff
@@ -292,7 +306,8 @@ def create_config(self):
 
         try:
             if not self.check_mode:
-                config_id = self.client.create_config(self.name, self.data, labels=labels)
+                config_id = self.client.create_config(self.name, self.data,
+                    labels=labels, templating=self.templating)
```
felixfontein, line 310:
```suggestion
                config_id = self.client.create_config(self.name, self.data, labels=labels, templating=self.templating)
```

felixfontein: Also this is a breaking change, since this will break every user not using Docker SDK for Python 5.0.3 (the latest version). That parameter has only been added there (https://docker-py.readthedocs.io/en/stable/change-log.html).

This way should work better:
```suggestion
                kwargs = {}
                if self.templating:
                    kwargs['templating'] = self.templating
                config_id = self.client.create_config(self.name, self.data, labels=labels, **kwargs)
```

Also you need to add `option_minimal_versions` to the `AnsibleDockerClient()` call, similar to https://github.com/ansible-collections/community.docker/blob/9a627f45c9480aebf669b67f21fad57f6023f560/plugins/modules/docker_swarm.py#L650-L673, where you specify that `template_driver` needs `docker_py_version='5.0.3', docker_api_version='1.37'` (https://docs.docker.com/engine/api/v1.37/#operation/ConfigList contains it, the API docs for 1.36 do not).

plugins/modules/docker_config.py
```diff
@@ -92,6 +98,7 @@
 author:
   - Chris Houseknecht (@chouseknecht)
   - John Hu (@ushuz)
+  - Sasha Jenner (@sashajenner)
```
felixfontein, line 101: Please note that by adding yourself here, you sign up as a maintainer of this module.

felixfontein (commented):
Thanks for your contribution! Please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).

plugins/modules/docker_config.py
sashajenner: I would rather not do that, thanks for letting me know.

sashajenner:
Thank you for the detailed review. I have made updates as a result. Please let me know what you think of these changes.

sashajenner:
It is possible to add some integration tests for this new option. Let me know if you think this is necessary or not.

felixfontein:
> It is possible to add some integration tests for this new option. Let me know if you think this is necessary or not.

This is a good idea! At least for the platforms which use the lastest Docker SDK for Python, and a new enough Docker daemon :) You simply need to add one or more tasks to tests/integration/targets/docker_config/tasks/test_docker_config.yml for that. Look at https://github.com/ansible-collections/community.docker/blob/main/tests/integration/targets/docker_container/tasks/tests/options.yml#L1046-L1084 for an example of a test which only works for certain Docker daemon and Docker SDK for Python versions: it ignores failures for all tasks, and then checks whether it behaved correctly.

felixfontein (commented):
I think `self.data_key` needs to be updated when templating is used. I'm not sure what's the best way to do this. First you should probably add a method `def _data_key_content(self):` which is used in line 255 as `self.data_key = hashlib.sha224(self._data_key_content()).hexdigest()`.

In that function, I would probably compile a configuration dictionary that's empty when `template` is not specified, and otherwise contains a key `template_driver` with the value of the template_driver option. If that config dict is empty, return `data`. Otherwise, return `data + SALT + json(config_dict)` (with sort_keys etc. so that JSON generation does not depend on platform), where `SALT` is a constant to be created containing a fixed but random string. (For example `nu7Zook2bo4Ohyah1ieghae2ohpho7Haa1yeefieNgoo9ov3que3yaiKeing1poh`.)

felixfontein:
Hmm, it would probably be better if you could determine for an existing config which templating method is used. If that is doable, use that to fix idempotence and leave `self.data_key` alone.

felixfontein:
You'll have to add the idempotence check in https://github.com/ansible-collections/community.docker/blob/26a61e98b8926b8ee6bffdd54a07e576c47f0150/plugins/modules/docker_config.py#L334-L347

sashajenner:
I think I've added an idempotency check now. It seems to work in testing. That is, when everything is the same but `template_driver` is modified, ansible says the task has changed.

However, I'm not confident in my solution. In particular, I'm not sure if I need to do a check as in https://github.com/ansible-collections/community.docker/blob/98b227003e93c4e736cb09d66dcd7f438935532f/plugins/modules/docker_config.py#L320 for `template_driver` as well.

plugins/modules/docker_config.py
```diff
@@ -316,17 +331,20 @@ def present(self):
...
             data_changed = False
+            template_driver_changed = False
             attrs = config.get('Spec', {})
             if attrs.get('Labels', {}).get('ansible_key'):
                 if attrs['Labels']['ansible_key'] != self.data_key:
                     data_changed = True
             else:
                 if not self.force:
                     self.client.module.warn("'ansible_key' label not found. Config will not be changed unless the force parameter is set to 'yes'")
+            if attrs['Labels']['template_driver'] != self.template_driver:
```
felixfontein, line 342: This should likely be something like:
```suggestion
            if attrs.get(...) != self.template_driver:
```
Unfortunately the API docs (https://docs.docker.com/engine/api/v1.41/#operation/ConfigInspect) do not give any hint how and where that information is available. Can you paste the output of something like `docker config inspect ...` for such a config?

In any case, this is one more reason why there should be tests for this in CI :)

sashajenner: For this docker config task
```
- name: debug
  community.docker.docker_config:
    name: db_password
    data: opensesame!
    state: present
    template_driver: golang
```
this output of `docker config inspect ...` looks like this
```
[root@447f901529a4 /]# docker config inspect db_password
[
    {
        "ID": "ilwv23ojrt2ooisz06enkmovj",
        "Version": {
            "Index": 28681
        },
        "CreatedAt": "2022-05-11T01:45:58.5768652Z",
        "UpdatedAt": "2022-05-11T01:45:58.5768652Z",
        "Spec": {
            "Name": "db_password",
            "Labels": {
                "ansible_key": "4c8ed891d2bcad5bd4359a985e82349ec42e0c04be5511b0e53ac258"
            },
            "Data": "b3BlbnNlc2FtZSE=",
            "Templating": {
                "Name": "golang"
            }
        }
    }
]
```
Hence, I now access the `attrs['Templating']['Name']` to get a config's `template_driver`. I have done some local testing and it seems to work. I will now work on adding integration tests.

felixfontein (approved):
LGTM

felixfontein:
@sashajenner thanks for contributing this!

PR #347, opened 2022-05-11
`docker_config` integration tests for `template_driver`

sashajenner:
##### SUMMARY
Adding integration tests for the `template_driver` option of `docker_config`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`template_driver` integration tests

### DIFF
--- a/tests/integration/targets/docker_config/tasks/test_docker_config.yml
+++ b/tests/integration/targets/docker_config/tasks/test_docker_config.yml
@@ -246,7 +246,7 @@
   - name: assert config was updated
     assert:
       that:
-         - output.changed
+         - output is changed
 
   - name: Invalid template_driver
     docker_config:
@@ -274,7 +274,7 @@
   - name: assert create config is idempotent
     assert:
       that:
-         - not output.changed
+         - output is not changed
 
   # data is the docker swarm's name
   - name: Update config with template data
@@ -288,7 +288,6 @@
   - name: Inspect config
     command: "docker config inspect {{ output.config_id }}"
     register: inspect
-    ignore_errors: yes
 
   - debug: var=inspect
 
--- a/tests/integration/targets/docker_config/tasks/test_docker_config.yml
+++ b/tests/integration/targets/docker_config/tasks/test_docker_config.yml
@@ -11,7 +11,7 @@
 
   - name: Create a Swarm cluster
     docker_swarm:
-      name: raw_eggplant
+      name: default
       state: present
       advertise_addr: "{{ansible_default_ipv4.address}}"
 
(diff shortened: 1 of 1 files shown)


sashajenner:
Seems to be failing since `templating` is not available for all versions.

felixfontein:
See here: https://github.com/ansible-collections/community.docker/pull/345#issuecomment-1116944929

tests/integration/targets/docker_config/tasks/test_docker_config.yml
```diff
@@ -226,6 +227,100 @@
...
+      name: db_password
+      data: opensesame!
+      template_driver: golang
+      state: present
+    register: output
+
+  - name: assert config was updated
+    assert:
+      that:
+         - output.changed
```
felixfontein, line 249:
```suggestion
         - output is changed
```

tests/integration/targets/docker_config/tasks/test_docker_config.yml
```diff
@@ -226,6 +227,100 @@
       that:
         - output.failed
 
+# template_driver tests
```
felixfontein, line 230: Put these into a `block:` with a `when:` condition that only matches when the feature is available.

tests/integration/targets/docker_config/tasks/test_docker_config.yml
```diff
@@ -226,6 +227,100 @@
...
+      name: db_password
+      data: opensesame!
+      template_driver: golang
+      state: present
+    register: output
+
+  - name: assert create config is idempotent
+    assert:
+      that:
+         - not output.changed
```
felixfontein, line 277:
```suggestion
         - output is not changed
```

tests/integration/targets/docker_config/tasks/test_docker_config.yml
```diff
@@ -226,6 +227,100 @@
...
+      name: db_password
+      data: "{{ '{{' }} .Service.Name {{ '}}' }}"
+      template_driver: golang
+      state: present
+    register: output
+
+  - name: Inspect config
+    command: "docker config inspect {{ output.config_id }}"
+    register: inspect
+    ignore_errors: yes
```
felixfontein, line 291: When the previous commands worked, this should not fail.
```suggestion
```

felixfontein (approved):

felixfontein:
@sashajenner thanks for providing these tests!

tests/integration/targets/docker_config/tasks/test_docker_config.yml
```diff
@@ -226,6 +227,99 @@
...
+    - name: Show inspection result
+      debug:
+        var: inspect
+
+    - name: assert config creation succeeded
+      assert:
+        that:
+           - "'db_password' in inspect.stdout"
+           - "'ansible_key' in inspect.stdout"
+           - "'\"Data\": \"default\"' in inspect.stdout"
```
felixfontein, line 304: BTW, this never passes, since (also confirmed by the API docs) `Data` is always the url-safe-Base64 encoded original data. So here you get `e3sgLlNlcnZpY2UuTmFtZSB9fQ==` and not `default`.

sashajenner: Right ok nice work

PR #354, opened 2022-05-18
docker_container: avoid breakage with Python 3.13

felixfontein:
##### SUMMARY
This fixes the currently failing Python 3.11 sanity tests (which were just added to ansible-test in devel branch).

The problem is that `pipes` is removed from Python 3.13's standard library. While looking at that, I noticed that we don't actually use that piece of code, so it can be removed trivially.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
docker_container


felixfontein:
CI failures are unrelated.

PR #368, opened 2022-05-21
CI: Replace the container used in the nsenter tests

felixfontein:
##### SUMMARY
CI currently fails because the quay.io/ansible/toolset:latest image no longer exists.

Let's try the one from https://github.com/ansible/python-base-image.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
nsenter connection


felixfontein:
Ok, we actually need ansible-playbook *in* the image, so let's try quay.io/ansible/ansible-runner:devel.

Issue #364, opened 2022-05-20
Future of community.docker


felixfontein:
community.docker has been using the [Docker SDK for Python](https://github.com/docker/docker-py/) ("docker-py") for almost everything (directly for most content, and indirectly for the docker_compose module; only the docker connection plugin and the docker_stack modules use the CLI tool).

Unfortunately, it looks like he Docker SDK for Python is pretty much dead ([PR for adding unmaintained notice](https://github.com/docker/docker-py/pull/2989), [activity since October 2021](https://github.com/docker/docker-py/graphs/contributors?from=2021-10-12&to=2022-05-31&type=c)). Maybe a fork will come up, maybe they'll allow some community folks to take the project over under its current repository and PyPi package name, and maybe it will just die (as the Python docker-compose did).

Anyway, this is a moment to think about what we could and/or should do. Having a fork or the original project continued would obviously be the best solutions ;-) With a fork we might have to adjust the import statements (resp. allow to import both original docker-py and the fork), but (if they don't do something drastic) most things can just stay as they are. In case the project is dead and noone wants to maintain a fork, we do have a problem.

We have four different groups of contents:

1. Content that uses the Docker CLI program, or something else:
   * `docker` connection plugin
   * `nsenter` connection plugin (uses the `nsenter` CLI tool)
   * `docker_machine` inventory plugin (uses the `docker-machine` CLI tool)
   * `current_container_facts` module (inspects files)
   * `docker_stack*` modules

2. Docker content that uses the API part of the Docker SDK for Python (i.e. not the high-level functionality)
   * `docker_api` connection plugin
   * `docker_containers` inventory plugin
   * `docker_swarm` inventory plugin
   * `docker_config` module
   * `docker_container_exec` module
   * `docker_container_info` module
   * `docker_container` module
   * `docker_host_info` module
   * `docker_image_info` module
   * `docker_image_load` module
   * `docker_image` module
   * `docker_login` module
   * `docker_network_info` module
   * `docker_network` module
   * `docker_node_info` module
   * `docker_node` module
   * `docker_prune` module
   * `docker_secret` module
   * `docker_swarm_info` module
   * `docker_swarm` module
   * `docker_swarm_service_info` module
   * `docker_swarm_service` module
   * `docker_volume_info` module
   * `docker_volume` module

3. Docker content that uses the high-level functionality of the Docker SDK for Python
   * `docker_plugin` module

4. Docker content that uses the Python version of docker-compose
   * `docker_compose` module

[...]

PR #373, opened 2022-05-24
Deprecate docker-compose EE dependency

felixfontein:
##### SUMMARY
We only just added it, but it already caused problems (see for example https://github.com/ansible/creator-ee/pull/68).

Maybe we need to do something similar for Docker SDK for Python as well, but first we need to figure out how to continue with #364.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
EE


markuman (approved):

felixfontein:
@markuman thanks for reviewing this!

PR #386, opened 2022-06-15
[WIP] Move PSF-license.txt to licenses/

felixfontein:
##### SUMMARY
Makes it easier to find all additional license files (next to COPYING).

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
licenses

PR #388, opened 2022-06-18
Deprecate Python 2.6

felixfontein:
##### SUMMARY
I think it's time to drop support for Python 2.6. Right now it is only supported with Docker SDK for Python < 2.0.0, and it will result in a lot of extra work for #364 and #386. If someone really has to use Python 2.6, they can happily continue using an older version of this collection, since the Docker SDK for Python they can use does not get updates anymore anyway.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
collection


gotmax23 (approved):
LGTM, but I'm not a member of the org.

PR #389, opened 2022-06-18
Improve/fix API version handling

felixfontein:
##### SUMMARY
Right now we retrieve the API version (the latest one supported by the daemon!) from the Docker daemon after creating the Docker SDK for Python API client. If the `version` module/plugin parameter is used, we should use that version instead, and if it is not used, the Python API client already retrieved that version for us. So let's safe one API call and use the API version supplied by the Python API client.

(The property exists since Docker SDK for Python 1.5.0: https://github.com/docker/docker-py/commit/7a6980d4794f3efaf85ad65458fbfac4c3afdec7 We require version 1.8.0 or newer, so we can depend on it.)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Docker SDK for Python based plugins and modules


briantist (approved):

felixfontein:

PR #387, opened 2022-06-18
[WILL BE SPLIT UP] Start adjusting some modules

felixfontein:
##### SUMMARY
Modules I want to adjust:

- [x] docker_container_exec
- [x] docker_container_info
- [ ] docker_container
- [x] docker_host_info
- [x] docker_image_info
- [x] docker_image_load
- [x] docker_image
- [x] docker_login
- [x] docker_network_info
- [x] docker_network
- [x] docker_prune
- [x] docker_volume_info
- [x] docker_volume

The idea is to support Python 2.7+. Maybe it also works with 2.6, but I don't know.

Ref: #364

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
various

PR #390, opened 2022-06-18
Move common utility code to plugins.module_utils.util

felixfontein:
##### SUMMARY
Subset of #387.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
various modules and plugins

### DIFF
--- a/changelogs/fragments/390-util.yml
+++ b/changelogs/fragments/390-util.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "Move common utility functions from the ``common`` module_util to a new module_util called ``util``. This should not have any user-visible effect (https://github.com/ansible-collections/community.docker/pull/390)."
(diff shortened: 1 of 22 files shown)


briantist (commented):
You may want to consider making the util internal so that "breaking" changes in the util interface that don't break things in the user-facing code (because it's changed at the same time), will not need to be a breaking change in the collection.

Details in this comment (it's your minimized comment, need to expand to see it):
- https://github.com/ansible/community/issues/539#issuecomment-780839686

(I did this for all of mine, since I don't expect or want any other collections to be depending on them)

briantist:
Ah, I've just realized that that stuff was mostly (completely?) moved from a different module_util, not from the modules themselves. I guess technically this is a breaking change then, though I'm personally not interested in being a stickler about that when we have a pretty good idea of when a module util is used internally only... I don't have such an idea on this one but I trust your judgement on it.

felixfontein:
It's not a breaking change because it's still available in the other module_utils (due import).

If you move a function (or class) `x` from module_utils `a` to module_utils `b` and add `from b import x` into `a`, `x` is now available from both.

briantist (approved):
I see, I missed that detail

PR #398, opened 2022-06-30
Vendor API connection code from Docker SDK for Python

felixfontein:
##### SUMMARY
Subset of #387: just add the vendored code and some module_utils and plugin_utils.

Once this is merged, I will create one PR per plugin/module from #387.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
module_utils
plugin_utils

### DIFF
--- a/changelogs/fragments/398-docker-api.yml
+++ b/changelogs/fragments/398-docker-api.yml
@@ -1,0 +1,5 @@
+major_changes:
+  - "The collection now contains vendored code from the Docker SDK for Python to talk to the Docker daemon.
+     Modules and plugins using this code no longer need the Docker SDK for Python installed on the machine
+     the module resp. plugin is running on
+     (https://github.com/ansible-collections/community.docker/pull/398)."
--- a/changelogs/fragments/398-docker-api.yml
+++ b/changelogs/fragments/398-docker-api.yml
@@ -1,5 +1,5 @@
 major_changes:
   - "The collection now contains vendored code from the Docker SDK for Python to talk to the Docker daemon.
      Modules and plugins using this code no longer need the Docker SDK for Python installed on the machine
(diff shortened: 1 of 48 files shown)


changelogs/fragments/398-docker-api.yml
```diff
@@ -0,0 +1,5 @@
+major_changes:
+  - "The collection now contains vendored code from the Docker SDK for Python to talk to the Docker daemon.
+     Modules and plugins using this code no longer need the Docker SDK for Python installed on the machine
+     the module resp. plugin is running on
```
briantist, line 4:
```suggestion
     the module or plugin is running on
```
nit: maybe this is just my brain after a long day but I don't understand the "resp." conjunction 😖

plugins/module_utils/common_api.py
```diff
@@ -0,0 +1,582 @@
...
+
+    def log(self, msg, pretty_print=False):
+        pass
+        # if self.debug:
+        #     log_file = open('docker.log', 'a')
+        #     if pretty_print:
+        #         log_file.write(json.dumps(msg, sort_keys=True, indent=4, separators=(',', ': ')))
+        #         log_file.write(u'\n')
+        #     else:
+        #         log_file.write(msg + u'\n')
```
briantist, line 139: Is this meant to stay in?

briantist (approved):
some very small things that are not worth withholding approval, nice work!

plugins/module_utils/common_api.py
felixfontein: This is copied from common.py, having it commented out makes it easy to comment it back in and have some basic logging on `debug=true`. So as long as it looks the same in common.py it should also be in common_api.py :)

felixfontein:

PR #414, opened 2022-07-02
Rewrite the docker_api connection plugin

felixfontein:
##### SUMMARY
Part of #387.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
docker_api

### DIFF
--- a/plugins/connection/docker_api.py
+++ b/plugins/connection/docker_api.py
@@ -12,8 +12,8 @@ short_description: Run tasks in docker containers
 version_added: 1.1.0
 description:
     - Run commands or put/fetch files to an existing docker container.
-    - Uses the Requests library to interact directly with the Docker daemon instead of
-      using the Docker CLI. Use the
+    - Uses the L(requests library,https://pypi.org/project/requests/) to interact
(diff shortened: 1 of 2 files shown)


plugins/connection/docker_api.py
```diff
@@ -12,7 +12,7 @@
 version_added: 1.1.0
 description:
     - Run commands or put/fetch files to an existing docker container.
-    - Uses Docker SDK for Python to interact directly with the Docker daemon instead of
+    - Uses the Requests library to interact directly with the Docker daemon instead of
```
briantist, line 15:
```suggestion
    - Uses the C(requests) library to interact directly with the Docker daemon instead of
```
this is totally opinion-based... feel free to use or ignore

briantist (approved):

plugins/connection/docker_api.py
felixfontein: I made it lower-case, but added a link in 31eaf86fc6456a59192048db63b64d780474c8d2.

briantist (approved):

PR #433, opened 2022-07-28
Prefer unitest.mock by universally using compat.mock

gotmax23:
##### SUMMARY

**Prefer unitest.mock by universally using compat.mock:**
`mock` is a backport of the `unittest.mock` module from the stdlib, and
there's no reason to use it on newer Python versions. `mock` is deprecated
in Fedora, so I figured I'd propose this here before downstream patching
our ansible-collection-community-docker package.

Note that there were some files that already had
```
try:
    from unittest import mock
except ImportError:
    import mock
```
but I modified them to use our compat module so all imports of mock are consistent.

**Sync compat.mock from latest ansible-core version:**
This removes old Python 3.4 compatibility code that is no longer needed.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests


gotmax23:
Does this need a changelog fragment even though it's not a user facing change? I would guess that it still does.

felixfontein:
> Does this need a changelog fragment even though it's not a user facing change? I would guess that it still does.

No, since this is a tests only change it should be fine.

tests/unit/compat/mock.py
```diff
@@ -1,109 +1,30 @@
...
-
-# Note: Could use the pypi mock library on python3.x as well as python2.x.  It
-# is the same as the python3 stdlib mock library
-
 try:
-    # Allow wildcard import because we really do want to import all of mock's
-    # symbols into this compat shim
-    # pylint: disable=wildcard-import,unused-wildcard-import
-    from unittest.mock import *
+    from unittest.mock import (
```
felixfontein, line 15: Why not keep the wildcard import here (and below)?

tests/unit/compat/mock.py
```diff
@@ -1,109 +1,30 @@
...
-        handle.read.side_effect = _read_side_effect
-        handle.readline.side_effect = _readline_side_effect()
-        handle.readlines.side_effect = _readlines_side_effect
+    from mock import (
+        call,
+        patch,
+        mock_open,
+        MagicMock,
+        Mock,
+    )
```
felixfontein, line 30:
```suggestion
```

tests/unit/compat/mock.py
gotmax23: This is what mattclay chose to do upstream. I could revert this part if you'd like.

felixfontein: I would keep it. In fact I would probably keep the old file, except the bottom part for old Python 3.x versions. (I.e. basically what community.internal_test_tools contains.)

tests/unit/plugins/module_utils/_api/api/test_client.py
```diff
@@ -36,14 +36,11 @@
 from ansible_collections.community.docker.plugins.module_utils._api.api.client import APIClient
 from ansible_collections.community.docker.plugins.module_utils._api.constants import DEFAULT_DOCKER_API_VERSION
 from ansible_collections.community.docker.plugins.module_utils._api.utils.utils import convert_filters
+from ansible_collections.community.docker.tests.unit.compat import mock
```
felixfontein, line 39: This cannot be imported here or in any of the following files, since this file is Apache licensed, while the imported module is GPLv3+. Same for the other files in tests/unit/plugins/module_utils/_api.

gotmax23: That is a bit of a stringent interpretation of the GPL, but fine. We might want to consider re-licensing it, though. In my opinion, this code is too trivial to really benefit from being GPL-licensed, anyways.

Also, it doesn't appear that having non-GPL test code is even [allowed](https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#id5). I think that rule should be changed, but that is currently the rule.

felixfontein: > That is a bit of a stringent interpretation of the GPL, but fine.

That's basically the reason why Ansible plugins must be licensed under GPLv3+, since they import code that is licensed under GPLv3+.

> Also, it doesn't appear that having non-GPL test code is even [allowed](https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst#id5). I think that rule should be changed, but that is currently the rule.

I think it's a grey area, since the statement "All other code must be under the GPL-3.0-or-later" is followed by "These plugins ...", which shows that this statement mainly applies to all other code **in `plugins/`**. This definitely should be clarified though.

felixfontein (approved):

felixfontein:
@gotmax23 thanks for this improvement!

tests/unit/plugins/module_utils/_api/api/test_client.py
gotmax23: > That's basically the reason why Ansible plugins must be licensed under GPLv3+, since they import code that is licensed under GPLv3+.

From my understanding, the license of the code in the repo doesn't really matter, as long as it's GPL-compatible. The GPL only applies once the code is distributed or compiled (in the case of compiled langauges). This project is already distributed under `GPL-3.0-or-later`. See e.g. https://github.com/ansible/ansible-lint#licensing.

felixfontein: Hmm, I'm no lawyer, so ... no idea :-) In any case, making sure that code importing GPLv3+ licensed code is itself GPLv3+ licensed code seems like a safeguard to me that will avoid potential legal problems that might or might not exist, so I'd rather stick to that :-)

gotmax23: > Hmm, I'm no lawyer, so ... no idea :-) In any case, making sure that code importing GPLv3+ licensed code is itself GPLv3+ licensed code seems like a safeguard to me that will avoid potential legal problems that might or might not exist, so I'd rather stick to that :-)

Seems reasonable to me.

PR #434, opened 2022-07-29
Vendored Docker SDK for Python updates

felixfontein:
##### SUMMARY
Incorporates new commits that apply from https://github.com/docker/docker-py/commits/main.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
vendored Docker SDK for Python components


felixfontein:
Current state contains commits up to https://github.com/docker/docker-py/commit/dff849f6bb7d6805f52dd2112b8666c86d3f3235.

changelogs/fragments/docker-py-changes-1.yml
```diff
@@ -0,0 +1,11 @@
...
+security_fixes:
+  - "modules and plugins communicating directly with the Docker daemon - when connecting by SSH and not using ``use_ssh_client=true``, reject unknown host keys instead of accepting them. This is only a change relatively to older community.docker 3.0.0 pre-releases, or with respect to Docker SDK for Python < 6.0.0. Docker SDK for Python 6.0.0 will also include this change (https://github.com/ansible-collections/community.docker/pull/434)."
+breaking_changes:
+  - "modules and plugins communicating directly with the Docker daemon - when connecting by SSH and not using ``use_ssh_client=true``, reject unknown host keys instead of accepting them. This is only a breaking change relatively to older community.docker 3.0.0 pre-releases, or with respect to Docker SDK for Python < 6.0.0. Docker SDK for Python 6.0.0 will also include this change (https://github.com/ansible-collections/community.docker/pull/434)."
```
gotmax23, line 11: Nitpick: `s/relatively/relative/'. I like adverbs (my name is Max**well**, after all), but this is not the correct place to use them. Perhaps it also makes sense to shorten `modules and plugins communicating directly with the Docker daemon` to just  `plugins communicating directly with the Docker daemon`, as modules are a type of plugin, but meh.

gotmax23 (approved):
I meant to submit that other comment as part of the review, but instead, I clicked the wrong button and got side tracked. All of the Python code looks good to me, and if we wanted to change it more, it would be nice to submit it upstream first. I would suggest preserving the entire commit messages from upstream in the future, instead of just the subjects, but I won't hold up the review on that.

changelogs/fragments/docker-py-changes-1.yml
felixfontein: I changed everything except the "modules and plugins ...". That because a) that formulation has been used in earlier changelog fragments, and b) many users do not know that modules are a special kind of plugin. Since the changelog is mainly for users, I think it's important to also mention modules here.

felixfontein:
> All of the Python code looks good to me, and if we wanted to change it more, it would be nice to submit it upstream first.

Upstream is Python 3.6+, while we also support Python 3.5 and Python 2.7. So I cannot copy the upstream changes literally, and they will not want the things we have to change for compatibility :-)

> I would suggest preserving the entire commit messages from upstream in the future, instead of just the subjects, but I won't hold up the review on that.

Good idea, I've included the entire messages for all commits (except the signed-off part, which I converted to `Co-authored-by:` so that GH shows correct attribution). The message is sometimes misleading (like mentioning that the code now uses `from e`, while we actually have to use six's `raise_from()`), but in general I think it's more helpful to have the full message included.

felixfontein:
@gotmax23 thanks for reviewing this!

PR #441, opened 2022-08-08
Fix docker_container tests

felixfontein:
##### SUMMARY
Something seems to be wrong. It only happens on RHEL 9.0 in CI though...

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
docker_container


felixfontein:
Looks like this is mainly a problem with the tests. I guess that for some reason the execution order of some tests changed, and hence the image that was there previously wasn't around anymore. That's why the task pulls the image - which results in the `changed` state -, which makes the test fail.

felixfontein:
c15dfa2bdef9981c0750160558e50f4f5ae24af9 solved the original issue, but now there's a new one (that already showed up in the re-runs).

PR #463, opened 2022-08-21
Documentation: changed cmd return value to command

1doce8:
##### SUMMARY
Seems like there is a slight issue. The cmd param doesn't exist, but command works well, also all examples use command param.

##### ISSUE TYPE
- Docs Pull Request

+label: docsite_pr

##### COMPONENT NAME
https://docs.ansible.com/ansible/latest/collections/community/docker/docker_compose_module.html#return-services/container_name/cmd

##### ADDITIONAL INFORMATION
If you try to specify cmd you will get the following error:
```
{"changed": false, "msg": "Configuration error - The Compose file '/tmp/ansible6ogp6awz/docker-compose.yml' is invalid because:\nUnsupported config option for services.envoy: 'cmd'"}
```

### DIFF
--- a/plugins/modules/docker_compose.py
+++ b/plugins/modules/docker_compose.py
@@ -339,7 +339,7 @@ services:
           returned: success
           type: complex
           contains:
-              cmd:
+              command:
                   description: One or more commands to be executed in the container.
                   returned: success
                   type: list


felixfontein (changes requested):
You are modifying the **return value documentation** here, but in the description and the error message you are quoting the **docker compose input** is referred to. So these might not be the same.

In fact, the integration tests prove (https://dev.azure.com/ansible/community.docker/_build/results?buildId=51786&view=logs&j=f22d7cc3-d298-538e-4fc5-8bdffa6eb6ef&t=5997e5e4-c1bb-5f00-496b-5e9d3a937295&l=613) that the return value is called `cmd`, and **not** `command`. So this change is invalid - at least for the version of docker-composed used in the tests (though I guess this should be the case for all 1.x versions).

PR #466, opened 2022-08-30
docker_swarm: add data_path_port option for swarm init

### LINKED ISSUE #296: Add support for "docker swarm init --data-path-port"
##### SUMMARY

The docker swarm init parameter `--data-path-port` is not supported, at least as I searched I did not find it.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`community.docker_swarm`

##### ADDITIONAL INFORMATION

`--data-path-port` can be passed to `docker swarm init` to use a different port for the bridge network. 

Background/use-case:
In our customers setup the port is was already used by another network in VMWare, so we were forced to set the bridge/VXLAN UDP port to a different one using mentioned parameter.

see https://docs.docker.com/engine/reference/commandline/swarm_init/#options

I would expect a YAML like this
```yaml
- name: Init a new swarm with different data-path-port parameter
  docker_swarm:
    state: present
    data_path_port: 4790
```

thehedgefrog:
##### SUMMARY

Allows usage of the `data_path_port` parameter when initializing a swarm.  Fixes #296.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

docker_swarm

##### ADDITIONAL INFORMATION

```yaml
- name: Init a new swarm with different data-path-port parameter
  docker_swarm:
    state: present
    data_path_port: 4790
```

### DIFF
--- a/changelogs/fragments/466-add-data-path-port.yml
+++ b/changelogs/fragments/466-add-data-path-port.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - docker_swarm - allows usage of the `data_path_port` parameter when initializing or joining a swarm (https://github.com/ansible-collections/community.docker/issues/296).
--- a/changelogs/fragments/466-add-data-path-port.yml
+++ b/changelogs/fragments/466-add-data-path-port.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - docker_swarm - allows usage of the ``data_path_port`` parameter when initializing or joining a swarm (https://github.com//issues/296).
+  - docker_swarm - allows usage of the ``data_path_port`` parameter when initializing a swarm (https://github.com/ansible-collections/community.docker/issues/296).
--- a/changelogs/fragments/466-add-data-path-port.yml
+++ b/changelogs/fragments/466-add-data-path-port.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - docker_swarm - allows usage of the `data_path_port` parameter when initializing or joining a swarm (https://github.com/ansible-collections/community.docker/issues/296).
+  - docker_swarm - allows usage of the ``data_path_port`` parameter when initializing or joining a swarm (https://github.com//issues/296).
--- a/plugins/modules/docker_swarm.py
+++ b/plugins/modules/docker_swarm.py
@@ -185,6 +185,13 @@ options:
         considered for idempotency checking.
     type: str
     version_added: 2.5.0
+  data_path_port:
+    description:
+      - Port to use for data path traffic.
+      - This needs to be a port number like C(9789).
+      - Only used when swarm is initialised or joined. Because of this it is not
+        considered for idempotency checking.
+    type: int
 extends_documentation_fragment:
 - community.docker.docker
 - community.docker.docker.docker_py_1_documentation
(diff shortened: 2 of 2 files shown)


changelogs/fragments/466-add-data-path-port.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - docker_swarm - allows usage of the `data_path_port` parameter when initializing or joining a swarm (https://github.com/ansible-collections/community.docker/issues/296).
```
felixfontein, line 2:
```suggestion
  - docker_swarm - allows usage of the ``data_path_port`` parameter when initializing or joining a swarm (https://github.com/ansible-collections/community.docker/issues/296).
```

plugins/modules/docker_swarm.py
```diff
@@ -185,6 +185,13 @@
         considered for idempotency checking.
     type: str
     version_added: 2.5.0
+  data_path_port:
+    description:
+      - Port to use for data path traffic.
+      - This needs to be a port number like C(9789).
+      - Only used when swarm is initialised or joined. Because of this it is not
+        considered for idempotency checking.
+    type: int
```
felixfontein, line 194:
```suggestion
    type: int
    version_added: 3.1.0
```

plugins/modules/docker_swarm.py
```diff
@@ -507,6 +521,7 @@ def init_swarm(self):
                 'advertise_addr': self.parameters.advertise_addr,
                 'listen_addr': self.parameters.listen_addr,
                 'data_path_addr': self.parameters.data_path_addr,
+                'data_path_port': self.parameters.data_path_port,
```
felixfontein, line 524: You have to do the same `if` dance as for `default_addr_pool` and `subnet_size` to add this one, since it is not supported by older Docker SDK for Python versions.

plugins/modules/docker_swarm.py
```diff
@@ -569,7 +584,8 @@ def join(self):
                 self.client.join_swarm(
                     remote_addrs=self.parameters.remote_addrs, join_token=self.parameters.join_token,
                     listen_addr=self.parameters.listen_addr, advertise_addr=self.parameters.advertise_addr,
-                    data_path_addr=self.parameters.data_path_addr)
+                    data_path_addr=self.parameters.data_path_addr,
+                    data_path_port=self.parameters.data_path_port)
```
felixfontein, line 588: Same here. You need to create a `kwargs = {}`, fill it with `data_path_port` if that's not `None`, and pass `**kwargs` here to avoid problems with older Docker SDK for Python versions.

felixfontein (commented):
Thanks for your contribution!

plugins/modules/docker_swarm.py
thehedgefrog: Discarded that as after more digging, data_path_port can only be used when initializing a swarm.

plugins/modules/docker_swarm.py
```diff
@@ -507,13 +522,16 @@ def init_swarm(self):
                 'advertise_addr': self.parameters.advertise_addr,
                 'listen_addr': self.parameters.listen_addr,
                 'data_path_addr': self.parameters.data_path_addr,
+                'data_path_port': self.parameters.data_path_port,
```
felixfontein, line 525: This line needs to be removed.

thehedgefrog: You're absolutely right, forgot to take it off.

felixfontein (approved):
LGTM. Will merge if nobody complains in a few days.

WojciechowskiPiotr:
LGTM

felixfontein:
@thehedgefrog thanks for improving this!

PR #500, opened 2022-11-22
Make image archive/save idempotent, using image id and repo tags as keys

iamjpotts:
##### SUMMARY

The `docker_image` component supports saving images to a tar file, similar to `docker image save` command. When a `docker_image` archive task is run repeatedly, it recreates the file every time.

After this change, the task will compare the Docker image ID of the archive file (if any) vs the image stored in the Docker Engine, and if they are the same, the archive file is left untouched.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`docker_image` used with `archive_path`

##### ADDITIONAL INFORMATION

Consider a scenario with two servers (`localhost` and `target`), where an Ansible playbook synchronizes images to a target server.

1. Playbook run against `localhost` uses `docker_load` with `archive_path` to export an image to a tar
2. Playbook run using `ansible.builtin.file` synchronizes the directory with the tar file on localhost with a target server directory
3. Playbook run using `docker_load` imports tar file on target server to target server's docker engine

Currently, this workflow regenerates the tar files over and over again, resending them to the target server each time.

After this change, Step 1 only exports the tar files once, until the source image changes, preventing the network copy from `localhost` to `target` on subsequent runs.
 
Export playbook

```
- hosts: localhost
  connection: local
  vars:
    docker_images:
      - nginx:1.20

  tasks:

    - name: Docker image export directory 
      ansible.builtin.file:
        path: ~/.docker-image-cache
        state: directory
        mode: '0755'

    - name: Export Docker Images
      community.docker.docker_image:
        name: nginx:1.20
        source: local
        archive_path: ~/.docker-image-cache/nginx-1.20.tar
        state: present
```

Import playbook

```

- hosts: target-server
  tasks:

    - name: Sync cache directory
      ansible.builtin.copy:
        src: ~/.docker-image-cache
        dest: ~/

    - name: Import Docker Images
      community.docker.docker_image:
        name: nginx:1.20
        source: load
        load_path: ~/.docker-image-cache/nginx-1.20.tar
        state: present

```

Output, without this PR, run 1 of 2.

```
PLAY [localhost] **************************************************************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Docker image export directory] ******************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Export Docker Images] ***************************************************************************************************************************************************************************************************************************************************
changed: [localhost]

PLAY RECAP ********************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=3    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   

```

Output, without this PR, run 2 of 2. Shows `changed` again

```

PLAY [localhost] **************************************************************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Docker image export directory] ******************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Export Docker Images] ***************************************************************************************************************************************************************************************************************************************************
changed: [localhost]

PLAY RECAP ********************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=3    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0 
```

Output, with this PR, run 1 of 2.

```
PLAY [localhost] **************************************************************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Docker image export directory] ******************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Export Docker Images] ***************************************************************************************************************************************************************************************************************************************************
changed: [localhost]

PLAY RECAP ********************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=3    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   

```

Output, with this PR, run 2 of 2. Shows `ok` on second run.

```

PLAY [localhost] **************************************************************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Docker image export directory] ******************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Export Docker Images] ***************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

PLAY RECAP ********************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=3    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0 
```

### DIFF
--- a/changelogs/fragments/500-idempotent-image-archival.yaml
+++ b/changelogs/fragments/500-idempotent-image-archival.yaml
@@ -1,0 +1,4 @@
+minor_changes:
+  - docker_image - when using ``archive_path``, detect whether changes are necessary based on the image ID (hash). If the existing tar archive matches the source, do nothing.
+    Previously, each task execution re-created the archive
+    (https://github.com/ansible-collections/community.docker/pull/500).
--- a/plugins/module_utils/image_archive.py
+++ b/plugins/module_utils/image_archive.py
@@ -1,0 +1,157 @@
+# Copyright 2022 Red Hat | Ansible
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import json
+import os
+import tarfile
+
+from ansible.module_utils.common.text.converters import to_native
+
+
+class ImageArchiveManifestSummary(object):
+    '''
+    Represents data extracted from a manifest.json found in the tar archive output of the
+    "docker image save some:tag > some.tar" command.
+    '''
+
+    def __init__(self, image_id, repo_tags):
+        '''
+        :param image_id:  File name portion of Config entry, e.g. abcde12345 from abcde12345.json
+        :type image_id: str
+        :param repo_tags  Docker image names, e.g. ["hello-world:latest"]
+        :type repo_tags: list
+        '''
+
+        self.image_id = image_id
+        self.repo_tags = repo_tags
+
+
+class ImageArchiveInvalidException(Exception):
+    def __init__(self, message, cause):
+        '''
+        :param message: Exception message
+        :type message: str
+        :param cause: Inner exception that this exception wraps
+        :type cause: Exception | None
+        '''
+
+        super(ImageArchiveInvalidException, self).__init__(message)
+
+        # Python 2 doesn't support causes
+        self.cause = cause
+
+
+def api_image_id(archive_image_id):
+    '''
+    Accepts an image hash in the format stored in manifest.json, and returns an equivalent identifier
+    that represents the same image hash, but in the format presented by the Docker Engine API.
+
+    :param archive_image_id: plain image hash
+    :type archive_image_id: str
+
+    :returns: Prefixed hash used by REST api
+    :rtype: str
+    '''
+
+    return 'sha256:%s' % archive_image_id
+
+
+def archived_image_manifest(archive_path):
+    '''
+    Attempts to get Image.Id and image name from metadata stored in the image
+    archive tar file.
+
+    The tar should contain a file "manifest.json" with an array with a single entry,
+    and the entry should have a Config field with the image ID in its file name, as
+    well as a RepoTags list, which typically has only one entry.
+
+    :raises:
+        ImageArchiveInvalidException: A file already exists at archive_path, but could not extract an image ID from it.
+
+    :param archive_path: Tar file to read
+    :type archive_path: str
+
+    :return: None, if no file at archive_path, or the extracted image ID, which will not have a sha256: prefix.
+    :rtype: ImageArchiveManifestSummary
+    '''
+
+    try:
+        # FileNotFoundError does not exist in Python 2
+        if not os.path.isfile(archive_path):
+            return None
+
+        tf = tarfile.open(archive_path, 'r')
+        try:
+            try:
+                ef = tf.extractfile('manifest.json')
+                try:
+                    text = ef.read().decode('utf-8')
+                    manifest = json.loads(text)
+                except Exception as exc:
+                    raise ImageArchiveInvalidException(
+                        "Failed to decode and deserialize manifest.json: %s" % to_native(exc),
+                        exc
+                    )
+                finally:
+                    # In Python 2.6, this does not have __exit__
+                    ef.close()
+
+                if len(manifest) != 1:
+                    raise ImageArchiveInvalidException(
+                        "Expected to have one entry in manifest.json but found %s" % len(manifest),
+                        None
+                    )
+
+                m0 = manifest[0]
+
+                try:
+                    config_file = m0['Config']
+                except KeyError as exc:
+                    raise ImageArchiveInvalidException(
+                        "Failed to get Config entry from manifest.json: %s" % to_native(exc),
+                        exc
+                    )
+
+                # Extracts hash without 'sha256:' prefix
+                try:
+                    # Strip off .json filename extension, leaving just the hash.
+                    image_id = os.path.splitext(config_file)[0]
+                except Exception as exc:
+                    raise ImageArchiveInvalidException(
+                        "Failed to extract image id from config file name %s: %s" % (config_file, to_native(exc)),
+                        exc
+                    )
+
+                try:
+                    repo_tags = m0['RepoTags']
+                except KeyError as exc:
+                    raise ImageArchiveInvalidException(
+                        "Failed to get RepoTags entry from manifest.json: %s" % to_native(exc),
+                        exc
+                    )
+
+                return ImageArchiveManifestSummary(
+                    image_id=image_id,
+                    repo_tags=repo_tags
+                )
+
+            except ImageArchiveInvalidException:
+                raise
+            except Exception as exc:
+                raise ImageArchiveInvalidException(
+                    "Failed to extract manifest.json from tar file %s: %s" % (archive_path, to_native(exc)),
+                    exc
+                )
+
+        finally:
+            # In Python 2.6, TarFile does not have __exit__
+            tf.close()
+
+    except ImageArchiveInvalidException:
+        raise
+    except Exception as exc:
+        raise ImageArchiveInvalidException("Failed to open tar file %s: %s" % (archive_path, to_native(exc)), exc)
--- a/plugins/modules/docker_image.py
+++ b/plugins/modules/docker_image.py
@@ -343,6 +343,13 @@ from ansible_collections.community.docker.plugins.module_utils.common_api import
     AnsibleDockerClient,
     RequestException,
 )
+
+from ansible_collections.community.docker.plugins.module_utils.image_archive import (
+    archived_image_manifest,
+    api_image_id,
+    ImageArchiveInvalidException,
+)
+
 from ansible_collections.community.docker.plugins.module_utils.util import (
     clean_dict_booleans_for_docker_api,
     DockerBaseClass,
@@ -535,6 +542,42 @@ class ImageManager(DockerBaseClass):
             self.results['actions'].append("Removed image %s" % (name))
             self.results['image']['state'] = 'Deleted'
 
+    @staticmethod
+    def archived_image_action(failure_logger, archive_path, current_image_name, current_image_id):
+        '''
+        If the archive is missing or requires replacement, return an action message.
+
+        :param failure_logger: a logging function that accepts one parameter of type str
+        :type failure_logger: Callable
+        :param archive_path: Filename to write archive to
+        :type archive_path: str
+        :param current_image_name: repo:tag
+        :type current_image_name: str
+        :param current_image_id: Hash, including hash type prefix such as "sha256:"
+        :type current_image_id: str
+
+        :returns: Either None, or an Ansible action message.
+        :rtype: str
+        '''
+
+        def build_msg(reason):
+            return 'Archived image %s to %s, %s' % (current_image_name, archive_path, reason)
+
+        try:
+            archived = archived_image_manifest(archive_path)
+        except ImageArchiveInvalidException as exc:
+            failure_logger('Unable to extract manifest summary from archive: %s' % to_native(exc))
+            return build_msg('overwriting an unreadable archive file')
+
+        if archived is None:
+            return build_msg('since none present')
+        elif current_image_id == api_image_id(archived.image_id) and [current_image_name] == archived.repo_tags:
+            return None
+        else:
+            name = ', '.join(archived.repo_tags)
+
+            return build_msg('overwriting archive with image %s named %s' % (archived.image_id, name))
+
     def archive_image(self, name, tag):
         '''
         Archive an image to a .tar file. Called when archive_path is passed.
@@ -559,9 +602,17 @@ class ImageManager(DockerBaseClass):
             self.log("archive image: image %s not found" % image_name)
             return
 
-        self.results['actions'].append('Archived image %s to %s' % (image_name, self.archive_path))
-        self.results['changed'] = True
-        if not self.check_mode:
+        # Will have a 'sha256:' prefix
+        image_id = image['Id']
+
+        action = self.archived_image_action(self.client.module.debug, self.archive_path, image_name, image_id)
+
+        if action:
+            self.results['actions'].append(action)
+
+        self.results['changed'] = action is not None
+
+        if (not self.check_mode) and self.results['changed']:
             self.log("Getting archive of image %s" % image_name)
             try:
                 saved_image = self.client._stream_raw_result(
@@ -579,8 +630,7 @@ class ImageManager(DockerBaseClass):
             except Exception as exc:
                 self.fail("Error writing image archive %s - %s" % (self.archive_path, to_native(exc)))
 
-        if image:
-            self.results['image'] = image
+        self.results['image'] = image
 
     def push_image(self, name, tag=None):
         '''
--- a/tests/integration/targets/docker_image/tasks/tests/options.yml
+++ b/tests/integration/targets/docker_image/tasks/tests/options.yml
@@ -7,10 +7,11 @@
   set_fact:
     iname: "{{ name_prefix ~ '-options' }}"
     iname_1: "{{ name_prefix ~ '-options-1' }}"
+    hello_world_alt: "{{ name_prefix }}-hello-world-alt:v1.2.3-foo"
 
 - name: Registering image name
   set_fact:
-    inames: "{{ inames + [iname, iname_1] }}"
+    inames: "{{ inames + [iname, iname_1, hello_world_alt] }}"
 
 ####################################################################
 ## build.args ######################################################
@@ -235,6 +236,63 @@
     source: pull
   register: archive_image
 
+- assert:
+    that:
+      - archive_image is changed
+
+- name: Copy archive because we will mutate it but other tests need the original
+  copy:
+    remote_src: true
+    src: "{{ remote_tmp_dir }}/image.tar"
+    dest: "{{ remote_tmp_dir }}/image_mutated.tar"
+
+- name: Archive image again (idempotent)
+  docker_image:
+    name: "{{ docker_test_image_hello_world }}"
+    archive_path: "{{ remote_tmp_dir }}/image_mutated.tar"
+    source: local
+  register: archive_image_2
+
+- assert:
+    that:
+      - archive_image_2 is not changed
+
+- name: Archive image 3rd time, should overwrite due to different id
+  docker_image:
+    name: "{{ docker_test_image_alpine_different }}"
+    archive_path: "{{ remote_tmp_dir }}/image_mutated.tar"
+    source: pull
+  register: archive_image_3
+
+- assert:
+    that:
+      - archive_image_3 is changed
+
+- name: Reset archive
+  copy:
+    remote_src: true
+    src: "{{ remote_tmp_dir }}/image.tar"
+    dest: "{{ remote_tmp_dir }}/image_mutated.tar"
+
+- name: Tag image with different name
+  docker_image:
+    name: "{{ docker_test_image_hello_world }}"
+    repository: "{{ hello_world_alt }}"
+    source: local
+
+- name: Archive image 4th time, should overwrite due to different name even when ID is same
+  docker_image:
+    name: "{{ hello_world_alt }}"
+    # Tagged as docker_test_image_hello_world but has same hash/id (before this task overwrites it)
+    archive_path: "{{ remote_tmp_dir }}/image_mutated.tar"
+    source: local
+  register: archive_image_4
+
+- assert:
+    that:
+      - archive_image_4 is changed
+
+# This is the test that needs the original, non-mutated archive
 - name: Archive image by ID
   docker_image:
     name: "{{ archive_image.image.Id }}"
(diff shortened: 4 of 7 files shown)


iamjpotts:
Had some difficulty running `ansible-test` locally; would like a maintainer to approve running the CI so I can compare its output vs local.

iamjpotts:
@felixfontein 

I've removed the 'do not merge' label.

I attempted to add two `.tar` files for use with unit tests but they got flagged by the license checker, since they are not text files and can't be checked for a copyright header.

I worked around that by base64-encoding the files and including them as a string literal.

The files were generated by me using `docker image save` - is this approach suitable for the licensing issue, or should another be considered?

felixfontein:
> I attempted to add two `.tar` files for use with unit tests but they got flagged by the license checker, since they are not text files and can't be checked for a copyright header.
> 
> I worked around that by base64-encoding the files and including them as a string literal.
> 
> The files were generated by me using `docker image save` - is this approach suitable for the licensing issue, or should another be considered?

I would really avoid adding binary blobs to this repository. How about creating these files during the integration tests (by running `docker image save`) and then using the freshly generated files?

iamjpotts:
One of the two tar files is very small (hello world is about 3kb). The other one is larger (about 3 mb) but I can find a smaller substitute, for example creating an image from `scratch` and adding a text file to it (e.g a copyright notice).

Would two small files be adequate? Is there a way to exclude them from release/distribution packaging?

The unit tests are easier and faster to run, and in principle, I think these tests belong there. They test parsing a file, as opposed to interacting with a service.

Regenerating them each time also introduces variability into the test input; it's nice to have the same exact input across all versions of Python being tested.

felixfontein:
> Would two small files be adequate? Is there a way to exclude them from release/distribution packaging?

I would **really** avoid adding any kind of binary file, if not absolutely necessary.

> The unit tests are easier and faster to run, and in principle, I think these tests belong there. They test parsing a file, as opposed to interacting with a service.

They are faster to run, but they also are a big problem, since you are including copyrighted contents into this repository. Both the hello-world executable and busybox come with copyright and legal conditions attached.

> Regenerating them each time also introduces variability into the test input; it's nice to have the same exact input across all versions of Python being tested.

I don't think this will be a problem.

plugins/modules/docker_image.py
```diff
@@ -879,6 +897,80 @@ def load_image(self):
...
+        if not os.path.isfile(archive_path):
+            return None
+
+        tf = tarfile.open(archive_path, 'r')
+        try:
+            try:
+                ef = tf.extractfile('manifest.json')
+
+                if hasattr(ef, '__exit__'):
+                    # Python 3
```
felixfontein, line 933: Better use `PY2` (`from ansible.module_utils.six import PY2`) to handle these differences.

plugins/modules/docker_image.py
```diff
@@ -879,6 +897,80 @@ def load_image(self):
...
+                )
+
+        finally:
+            # In Python 2.6, TarFile does not have exit
+            if hasattr(tf, '__exit__'):
+                with tf:
+                    pass
+
+    except Exception as exc:
+        raise ImageArchiveInvalidException("Failed to open tar file %s: %s" % (archive_path, to_native(exc)))
```
felixfontein, line 971: Same here.

.gitignore
```diff
@@ -134,3 +134,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# PyCharm
+.idea
```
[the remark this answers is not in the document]
iamjpotts: Done, see #501

plugins/modules/docker_image.py
```diff
@@ -569,8 +587,8 @@ def archive_image(self, name, tag):
             except Exception as exc:
                 self.fail("Error writing image archive %s - %s" % (self.archive_path, to_native(exc)))
 
-        if image:
-            self.results['image'] = image
+        self.results['old_image_id'] = archived_id
+        self.results['image'] = image
```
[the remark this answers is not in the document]
iamjpotts: Earlier in the function, there's already a guard on `image` with a `return` to exit the function.

https://github.com/ansible-collections/community.docker/blob/main/plugins/modules/docker_image.py#L548

plugins/modules/docker_image.py
```diff
@@ -569,8 +587,8 @@ def archive_image(self, name, tag):
             except Exception as exc:
                 self.fail("Error writing image archive %s - %s" % (self.archive_path, to_native(exc)))
 
-        if image:
-            self.results['image'] = image
+        self.results['old_image_id'] = archived_id
```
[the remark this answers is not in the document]
iamjpotts: I'll delete it to simplify the PR.

plugins/modules/docker_image.py
```diff
@@ -549,9 +550,26 @@ def archive_image(self, name, tag):
...
             return
 
+        # Will have a 'sha256:' prefix
+        new_image_id = image['Id']
+
+        try:
+            archived_id = archived_image_id(self.archive_path)
+
+            if archived_id:
+                self.results['actions'].append('Previous image ID was %s' % archived_id)
```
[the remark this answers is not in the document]
iamjpotts: OK.

What are the differences between `self.log` and `module.debug` ?

I have no opinion either way, but want to understand them better.

plugins/modules/docker_image.py
```diff
@@ -879,6 +897,80 @@ def load_image(self):
...
+                # Returns hash without 'sha256:' prefix
+                try:
+                    # Strip off .json filename extension, leaving just the hash.
+                    return os.path.splitext(config_file)[0]
+                except Exception as exc:
+                    raise ImageArchiveInvalidException(
+                        "Failed to extract image id from config file name %s: %s" % (config_file, to_native(exc))
+                    )
+            except Exception as exc:
+                raise ImageArchiveInvalidException(
```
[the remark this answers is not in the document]
iamjpotts: Ugh, my exception handling looks awful, I should have stopped before 4 am.

iamjpotts:
Changes:

* Added change log fragment
* Replaced committed base-64 encoded tar files with runtime-generated stubs with same structure
* Fixed nested exception handling and added tests for causes
* Replaced action entries with `log.debug` statements (`self.log` calls an empty function with `pass` inside)
* Replaced `__exit__` attribute conditionals with `close()` calls
* Removed `old_image_id` result

changelogs/fragments/500-idempotent-image-archival.yaml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - docker_image - archive_path - based on image id (hash), if the existing tar archive matches the source, do nothing.
```
felixfontein, line 2:
```suggestion
  - docker_image - when using ``archive_path``, detect whether changes are necessary based on the image ID (hash). If the existing tar archive matches the source, do nothing.
```

changelogs/fragments/500-idempotent-image-archival.yaml
```diff
@@ -0,0 +1,3 @@
...
+    Previously, each task execution re-created the archive.
```
felixfontein, line 3:
```suggestion
    Previously, each task execution re-created the archive
    (https://github.com/ansible-collections/community.docker/pull/500).
```

plugins/modules/docker_image.py
```diff
@@ -370,6 +375,9 @@
 )
 
 
+log = logging.getLogger(__name__)
```
felixfontein, line 378: Please remove this.

plugins/modules/docker_image.py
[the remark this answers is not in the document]
felixfontein: `self.log()` is the preferred way for the Docker modules, since all other Docker modules also use this. It does nothing, but can easily be enable to print something when debugging a module in the collection.

plugins/modules/docker_image.py
```diff
@@ -549,9 +557,26 @@ def archive_image(self, name, tag):
...
             return
 
+        # Will have a 'sha256:' prefix
+        new_image_id = image['Id']
+
+        try:
+            archived_id = archived_image_id(self.archive_path)
+
+            if archived_id:
+                log.debug('Previous image ID was %s', archived_id)
```
felixfontein, line 567: Please use `self.log()` or - if really necessary - `module.debug()`.

plugins/modules/docker_image.py
```diff
@@ -549,9 +557,26 @@ def archive_image(self, name, tag):
...
+            else:
+                log.debug('No previous image')
+        except ImageArchiveInvalidException:
+            log.debug('Previous image ID not available', exc_info=True)
+            archived_id = None
+
+        changed = (new_image_id != ('sha256:%s' % archived_id))
+
         self.results['actions'].append('Archived image %s to %s' % (image_name, self.archive_path))
```
felixfontein, line 576: This only needs to be added if `changed` is `True`. Otherwise, the module is not doing something, and adding this action would then be wrong.

felixfontein:
You have a very strange commit list, I guess you should probably rebase to get rid of all the stray changes which do not belong into this PR.

plugins/module_utils/image_archive.py
```diff
@@ -0,0 +1,130 @@
...
+def api_image_id(archive_image_id):
+    """
+    Accepts an image hash in the format stored in manifest.json, and returns an equivalent identifier
+    that represents the same image hash, but in the format presented by the Docker Engine API.
+    """
+
+    return 'sha256:%s' % archive_image_id
+
+
+def archived_image_action(failure_logger, archive_path, current_image_name, current_image_id):
```
felixfontein, line 33: I don't think this function belongs in this file. It should be part of the docker_image module.

felixfontein:
PS: The `[DO NOT MERGE]` prefix is for PRs that will never be merged and are only for testing something. I've marked your PR as a draft, if you want to use a prefix use `[WIP]` or `WIP:`.

plugins/module_utils/image_archive.py
iamjpotts: Heh I thought about that and waffled on it.

You prefer a standalone function or a method?

iamjpotts:
> You have a very strange commit list, I guess you should probably rebase to get rid of all the stray changes which do not belong into this PR.

Yeah it's screwed up bc of when I rebased vs when the source main `branch` was updated yesterday. I'll rebase again soon.

plugins/module_utils/image_archive.py
felixfontein: Considering that most functionality in that module is a method, I would make it a method there.

iamjpotts: There's some shared functions for stubbing tar files that would be needed by both tests in `modules/` and `module_utils/` and I could create a conftest.py in their parent to share them, but that scopes them across all tests, as opposed to scoping them to just `docker_image`.

Is that OK?

If I leave `def archived_image_action` where it is, the new `conftest.py` isn't needed at all.

felixfontein: You can also put unit test code in arbitrary `.py` files and just import it. There's no need to always use `conftest.py` for that.

plugins/modules/docker_image.py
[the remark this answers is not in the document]
iamjpotts: [This line](https://github.com/ansible-collections/community.docker/pull/500/files#diff-6178e5dc28781ab3227b344e50384dafb10d5df4aa0b54dc6fa3585c5c259d14R549) is the only thing logged now:

It logs an exception, if any, from reading or parsing the tar file:

```
'Unable to extract manifest summary from archive: %s' % to_native(exc)
```

How do I call `module.debug` ?

That doesn't resolve in my IDE.

plugins/modules/docker_image.py
iamjpotts: Removed.

plugins/modules/docker_image.py
iamjpotts: Changed to self.log, but would like to know how to call `module.debug` to report the exception.

plugins/modules/docker_image.py
iamjpotts: Done.

plugins/module_utils/image_archive.py
iamjpotts: Added a new file `tests/unit/plugins/test_support/docker_image_archive_stubbing.py`

iamjpotts:
I refactored `archived_image_action` onto `ImageManager` but stopped short of making it an instance method due to all the bootstrapping / mocking that would be required.

`ImageManager` has several "sub tasks" - archival being one of them - that would be easier to test if they were composed from separate classes, but that would be way too invasive of a refactor here.

tests/integration/targets/docker_image/tasks/tests/options.yml
```diff
@@ -3,6 +3,9 @@
 # GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
+- name: Include docker_image test variables
+  include_vars: '../../vars/main.yml'
+
```
felixfontein, line 8:
```suggestion
```
`vars/main.yml` is automatically loaded and made available in a role.

tests/integration/targets/docker_image/tasks/tests/options.yml
```diff
@@ -235,6 +238,63 @@
...
+- name: Reset archive
+  copy:
+    remote_src: true
+    src: "{{ remote_tmp_dir }}/image.tar"
+    dest: "{{ remote_tmp_dir }}/image_mutated.tar"
+
+- name: Tag image with different name
+  docker_image:
+    name: "{{ docker_test_image_hello_world }}"
+    repository: "{{ docker_test_image_hello_world_archive_alt }}"
```
felixfontein, line 282: You need to register this image name in line 16 (`inames: "{{ inames + [iname, iname_1, docker_test_image_hello_world_archive_alt] }}"`).

tests/integration/targets/docker_image/vars/main.yml
```diff
@@ -0,0 +1,6 @@
+# Copyright 2022 Red Hat | Ansible
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# Used to create an alternate tag of an image and verify the archive will re-generate even with same image hash/id.
+docker_test_image_hello_world_archive_alt: docker_test_image_hello_world:latest
```
felixfontein, line 6: Why use a fixed name and not generate one (like `iname` and `iname_1`)? That way the name should be unique per test, and potential leftovers from previous runs do not cause problems.

tests/unit/plugins/modules/test_docker_image.py
```diff
@@ -0,0 +1,119 @@
...
+    write_imitation_archive,
+    write_irrelevant_tar,
+)
+
+
+def assert_no_logging(msg):
+    raise AssertionError('Should not have logged anything but logged %s' % msg)
+
+
+def assert_log(messages):
```
felixfontein, line 27: The name of this function is very generic; it should be more specific since this function tests for a very specific input.

tests/unit/plugins/test_support/docker_image_archive_stubbing.py
```diff
@@ -0,0 +1,81 @@
...
+            'RepoTags': repo_tags
+        }
+    ]
+
+    write_imitation_archive_with_manifest(file_name, manifest)
+
+
+def write_imitation_archive_with_manifest(file_name, manifest):
+    manifest_file_name = tempfile.mkstemp(suffix='json')[1]
+
```
felixfontein, line 41: Shouldn't the `try:` start here, since the temp file has already been created and can be removed if the following write fails for whatever reason?

tests/unit/plugins/test_support/docker_image_archive_stubbing.py
```diff
@@ -0,0 +1,81 @@
...
+            tf.close()
+
+    finally:
+        os.remove(manifest_file_name)
+
+
+def write_irrelevant_tar(file_name):
+    inner_file_name = tempfile.mkstemp(suffix='txt')[1]
+
+    with open(inner_file_name, 'w') as f:
```
felixfontein, line 65: Same here.

tests/integration/targets/docker_image/tasks/tests/options.yml
iamjpotts: Why did the tests work without doing this, and what's gained by adding it here?

Is this required due to removing the new `include_vars` task from the top of `options.yml` ?

tests/integration/targets/docker_image/vars/main.yml
iamjpotts: The other examples I saw used fixed names, e.g. `tests/integration/targets/setup_docker_registry/vars/main.yml` (granted, those are referencing external images).

How do I auto-generate a name?

tests/unit/plugins/modules/test_docker_image.py
iamjpotts: Ironically that function is not used after the refactoring so I'm deleting it.

tests/unit/plugins/test_support/docker_image_archive_stubbing.py
iamjpotts: Yes, good catch.

I also moved the inner file management inside of the tar file `try..finally`

tests/unit/plugins/test_support/docker_image_archive_stubbing.py
iamjpotts: Another great catch, made same two changes (your suggestion + moving inner file create/delete inside of tar file try..finally)

tests/integration/targets/docker_image/tasks/tests/options.yml
iamjpotts: nvm answered my own question by trying it lol

tests/integration/targets/docker_image/tasks/tests/options.yml
iamjpotts: Done.

plugins/modules/docker_image.py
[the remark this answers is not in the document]
felixfontein: Did you look in https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1305 ?

tests/integration/targets/docker_image/vars/main.yml
felixfontein: Check out lines 8 and 9 in that file. If you want extra randomness add something like `(10000 | random)`.

iamjpotts: I moved it. I had to add an explict tag suffix because some tasks were interpreting it using `:latest` and others interpreted it using `:hello-world`

iamjpotts:
Made one more docstring change to tickle the build after a transient network related failure.

Ready for review.

plugins/modules/docker_image.py
```diff
@@ -339,10 +339,15 @@
 
 from ansible.module_utils.common.text.converters import to_native
 
-from ansible_collections.community.docker.plugins.module_utils.common_api import (
+from ..module_utils.common_api import (
```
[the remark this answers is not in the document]
iamjpotts: Reverted.

There's problems with the imports that using a relative import fixes.

I'll open a separate issue.

plugins/plugin_utils/common_api.py
```diff
@@ -9,7 +9,7 @@
 from ansible.errors import AnsibleConnectionFailure
 from ansible.utils.display import Display
 
-from ansible_collections.community.docker.plugins.module_utils.common_api import (
+from ..module_utils.common_api import (
```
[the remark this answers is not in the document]
iamjpotts: Done.

plugins/modules/docker_image.py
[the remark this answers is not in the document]
iamjpotts: #507

plugins/modules/docker_image.py
```diff
@@ -343,6 +343,11 @@
     AnsibleDockerClient,
     RequestException,
 )
+from ..module_utils.image_archive import (
```
felixfontein, line 346:
```suggestion
from ansible_collections.community.docker.plugins.module_utils.image_archive import (
```

iamjpotts:
I'm now getting build failures which are timeouts waiting for the temporary docker registry to start.

[For example](https://dev.azure.com/ansible/community.docker/_build/results?buildId=61726&view=logs&j=7eeace52-ac01-56c9-ddb0-19dd66139108&t=2915ff0d-5d52-54f6-b188-708c94614a79)

Made a whitespace change to tickle the build and it failed the same way (three times so far).

I don't _think_ this is related to accepting your import change suggestion, but I rebased right away.

Looking at the CI log, I suspect the registry container is failing to stay running (the previous successful task is a start request), but there isn't any log output from the registry container in the CI log to see why.

felixfontein:
This might be caused by https://github.com/ansible/ansible/pull/78550. Simply ignore that failure for now.

felixfontein:
Now there's a conflict due to the other PR being merged. After rebasing with the latest `main` branch all tests should also pass.

felixfontein (approved):

iamjpotts:
> Now there's a conflict due to the other PR being merged. After rebasing with the latest `main` branch all tests should also pass.

I'm updating the docstring style in this PR also, it's not committed yet. I'll drop a message here once the build is green and I've looked over the code one last time.

felixfontein (approved):

iamjpotts:
Build is green.

I did miss one triple-quote to single quote in a [test_docker_image.py](https://github.com/ansible-collections/community.docker/pull/500/files#diff-f918e6be7fd2af82f28f70c43902120ed64d6460ae46c8a6957b9a8ad5e67f83R35)

If that doesn't concern you, I don't see anything else to change.

felixfontein:
@iamjpotts thanks a lot for your contribution!

PR #510, opened 2022-11-30
Make current_container_facts work with newer Docker versions and latest ansible-test container changes

felixfontein:
##### SUMMARY
It does seem to no longer work, might be related to https://github.com/ansible/ansible/pull/78550.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
current_container_facts

### DIFF
--- a/plugins/modules/current_container_facts.py
+++ b/plugins/modules/current_container_facts.py
@@ -113,7 +113,7 @@ def main():
         for line in contents.splitlines():
             parts = line.split()
             if len(parts) >= 5 and parts[4] == '/etc/hostname':
-                m = re.match('/var/lib/docker/containers/([a-f0-9]+)/hostname', parts[3])
+                m = re.match('.*/docker/containers/([a-f0-9]+)/hostname', parts[3])
                 if m:
                     container_id = m.group(1)
                     container_type = 'docker'
--- a/plugins/modules/current_container_facts.py
+++ b/plugins/modules/current_container_facts.py
@@ -54,9 +54,9 @@ ansible_facts:
         ansible_module_container_type:
             description:
               - The detected container environment.
-              - Contains an empty string if no container was detected.
-              - Otherwise, will be one of C(docker), C(azure_pipelines), or C(github_actions).
+              - Contains an empty string if no container was detected, or a non-empty string identifying the container environment.
               - C(github_actions) is supported since community.docker 2.4.0.
+              - C(podman) is supported since community.docker 3.3.0.
             returned: always
             type: str
             choices:
@@ -64,9 +64,11 @@ ansible_facts:
               - docker
               - azure_pipelines
               - github_actions
+              - podman
 '''
 
 import os
+import re
 
 from ansible.module_utils.basic import AnsibleModule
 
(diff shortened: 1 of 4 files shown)


felixfontein:
CC @mattclay, looking at /proc/self/cpuset to figure out the current container breaks down on newer Docker daemons, and also no longer works with the `alpine3` container on AZP in ansible-test since https://github.com/ansible/ansible/pull/78550 was merged (don't ask me why only for that one...). I did some trying around, and looking at `/proc/self/mountinfo` and searching for `/etc/hostname` seems to work pretty well as a quick replacement (though that assumes that the user didn't change their docker data directory to something else than `/var/lib/docker/` - or maybe this is also a fixed string that isn't related to the position of that directory on the host system?). This might be useful for https://github.com/ansible/ansible/blob/devel/test/lib/ansible_test/_internal/docker_util.py#L576.

plugins/modules/current_container_facts.py
```diff
@@ -54,37 +54,42 @@
...
+    mountinfo_path = '/proc/self/mountinfo'
+
     container_id = ''
     container_type = ''
 
-    if os.path.exists(path):
+    contents = None
+    if os.path.exists(cpuset_path):
         # File content varies based on the environment:
         #   No Container: /
```
kristianheljas, line 88: This comment is not reflecting the changes. But i'm hesitant whether it should be updated or removed.

plugins/modules/current_container_facts.py
```diff
@@ -101,6 +106,23 @@ def main():
...
             container_type = 'github_actions'
 
+    if not container_id and os.path.exists(mountinfo_path):
+        with open(mountinfo_path, 'rb') as f:
+            contents = f.read().decode('utf-8')
+
+        for line in contents.splitlines():
+            parts = line.split()
+            if len(parts) >= 5 and parts[4] == '/etc/hostname':
+                m = re.match('/var/lib/docker/containers/([a-f0-9]+)/hostname', parts[3])
```
kristianheljas, line 116: Testing on docker version 20.10.21, it starts with `/@/var/...`
I'm using Arch Linux (kernel 6.0.10-arch2-1) with cgroups v2 if that matters

kristianheljas (changes requested):
Other than the outdated comment and test with docker, look good. Podman worked flawlessly

plugins/modules/current_container_facts.py
kristianheljas: https://gist.github.com/kristianheljas/b20df29c43ab96fddbd7acd36d45a965

kristianheljas: Oh, the `/@` is btrfs subvolume - silly me! This can probably be longer as well.
Matching only the end of the should still be strict enough to account for that.

mattclay:
@felixfontein Looking at `/proc/self/cpuset` is unreliable because of cgroup namespaces. It contains a host-specific path only when the host namespace is used:

```shell
$ docker run -it --rm --cgroupns host quay.io/bedrock/alpine:3.17.0 cat /proc/self/cpuset
/docker/a34f76a8f390f33188505c9be17734fcd41d6b1afc014bb36ad0b030a24ffe11
$ docker run -it --rm --cgroupns private quay.io/bedrock/alpine:3.17.0 cat /proc/self/cpuset
/
```

plugins/modules/current_container_facts.py
felixfontein: Implemented in 34e94e078c7d1624475d9f95700bdfdd545250ef.

kristianheljas: /resolved

plugins/modules/current_container_facts.py
felixfontein: I extended the comment in 23a07989327966b734d26a8521e3c8698b3595d0, I hope my extension is accurate enough :)

kristianheljas: /resolved

felixfontein:
@kristianheljas @mattclay thanks for testing, reviewing, and commenting!

felixfontein:
Manual backport of bugfix part to stable-2 in #512.

PR #531, opened 2022-12-11
CI: add CentOS Stream 8 with Python 3.6 to matrix

felixfontein:
##### SUMMARY
Uses VMs for Alpine 3.16, Fedora 36, Ubuntu 20.04, and Ubuntu 22.04, instead of Docker containers.

This increases the size of the CI matrix since for these platfoms, the test groups 1-3 have not been running before. (These are for Docker Swarm tests which need to reconfigure the Docker daemon, and thus are only run on VMs. Which so far was only RHEL.)

CC @mattclay (as I don't know whether these VMs are supposed to be used by collections as well)

CC @Spredzy @Andersson007 as this extends the CI matrix and RH is paying for that :)

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -111,17 +111,16 @@ stages:
           targets:
             - name: CentOS 7
               test: centos7
+            - name: Fedora 36
+              test: fedora36
             - name: openSUSE 15
               test: opensuse15
-            # The following are already tested as VMs with devel:
-            # - name: Fedora 36
-            #   test: fedora36
-            # - name: Ubuntu 20.04
-            #   test: ubuntu2004
-            # - name: Ubuntu 22.04
-            #   test: ubuntu2204
-            # - name: Alpine 3
-            #   test: alpine3
+            - name: Ubuntu 20.04
+              test: ubuntu2004
+            - name: Ubuntu 22.04
+              test: ubuntu2204
+            - name: Alpine 3
+              test: alpine3
           groups:
             - 4
             - 5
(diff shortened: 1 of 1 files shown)


felixfontein:
Ah, I forgot to mention that this also adds two additional CI runs with CentOS Stream 8 (in a Docker container) with Python 3.6 (the system Python on that OS) next to the CentOS Stream 8 tests with Python 3.9.

mattclay:
See my related comment: https://github.com/ansible-collections/community.crypto/pull/541#issuecomment-1346934179 

Given the types of tests needed, I think it makes more sense to utilize VMs than it does for most other collections. However, we should still try to limit VM usage to the tests that won't work in containers.

felixfontein:
Actually I'm not sure how much different VMs will help here. The main differences probably come from different Docker versions rather than from different OSes (and of course different Python versions and different versions of Docker SDK for Python, for the modules that still use that).

Right now the modules that need VMs for tests are the modules that require Docker Swarm, since reconfiguring the Docker daemon (enable and disable Swarm mode) is rather problematic especially when the tests themselves run in containers (I'm not sure whether that actually worked in the past, or whether we always resticted it to VMs... too long ago...). Also for Docker Swarm I'm really wondering how actively it is still used, and how much CI resources should really be used for it.

So if nobody has convincing arguments for why we need to run the Docker Swarm tests on more platforms, I'll probably restrict this PR to only add CentOS Steam 8 with Python 3.6 and skip the additional VMs. (And if we ever encounter a situation we really need VMs we can still reconsider.)

felixfontein:
This PR is now only adding one single new CI run (CentOS Stream 8 with system Python 3.6).

felixfontein:
@mattclay thanks for your comments!

Issue #216, opened 2021-10-01
docker-compose 2.0.0 is no longer written in Python, thus breaking the docker_compose module


felixfontein:
##### SUMMARY
I guess we have to mention in the requirements of the module that it needs docker-compose < 2.0.0. Potentially we either have to adjust the module to also work with docker-compose 2.0.0 and newer, or we have to deprecate it.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
docker_compose

felixfontein:
CC @aminvakil thanks for reporting this! (BTW the PR you mentioned is unrelated to this.)

felixfontein:
Adding a warning and the < 2.0.0 requirement in #217.

dmp1ce:
Should I expect that this module will be updated to support version 2.0.0 and greater of Compose? I'm trying to figure out if I should rewrite my docker-compose sections of my playbook or just wait for an update.

felixfontein:
@dmp1ce that depends on whether someone wants to rewrite the docker_compose module in such a way that it will simply work with the CLI. Right now the module depends on importing the Python code of docker-compose itself, which only works for docker-compose < 2.0.0 since for 2.0.0 docker-compose was rewritten in Go.

libricoleur:
This has become a problem on distros that already updated to compose v2, such as Arch Linux. Has anyone started working on a port of the module?

In the meantime I installed v1 from pip in a virtualenv and pointed the PYTHONPATH environment variable to it.

felixfontein:
I'm not aware of anyone working on this.

dmp1ce:
@libricoleur I just created my own tasks which copy the `docker-compose.yml` to the machine and then run `docker-compose`. It looks something like this:

```

- name: Loop through starting Docker services (manually)
  block:
    - name: Copy docker-compose templates for each service
      ansible.builtin.template:
        src: files/docker_services/{{ item.project_name }}.yml.j2
        dest: /home/maintenance/docker_services/{{ item.project_name }}/docker-compose.yml
        backup: yes
      loop: "{{ docker_services }}"

    - name: "Update docker service image"
      command:
        chdir: /home/maintenance/docker_services/{{ item.project_name }}
        cmd: docker-compose pull
      loop: "{{ docker_services }}"
    - name: "Start docker service"
      command:
        chdir: /home/maintenance/docker_services/{{ item.project_name }}
        cmd: docker-compose up -d
      loop: "{{ docker_services }}"
[...]

PR #540, opened 2022-12-29
Add docker compose v2 module

Sid-Sun:
Signed-off-by: Sid Sun <sid@dronk.dev>

##### SUMMARY
Add module to support docker compose v2 via python-on-whales 
the docker_compose_v2 module does not behave the same as the existing module, it lacks some features and does some things differently
returns from module are somewhat finicky as [python on whales has a slightly weird logic to it](https://github.com/gabrieldemarmiesse/python-on-whales/issues/395) 

adds lacking support for compose v2 #216

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
docker_compose_v2

### DIFF
--- a/plugins/modules/docker_compose_v2.py
+++ b/plugins/modules/docker_compose_v2.py
@@ -1,0 +1,371 @@
+#!/usr/bin/python
+
+from __future__ import (absolute_import, division, print_function)
+from ansible.module_utils.basic import AnsibleModule
+import os
+from python_on_whales.docker_client import DockerClient
+from python_on_whales.exceptions import (
+    DockerException,
+)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: docker_compose_v2
+
+short_description: Manage multi-container Docker applications with Docker Compose v2.
+
+version_added: "1.0.0"
+
+description:
+  - Uses Docker Compose v2 to manage docker services. B(This module requires docker-compose > 2.0.0.)
+  - Configuration can be read from a C(docker-compose.yaml) file.
+  - See the examples for more details.
+  - Supports check mode.
+
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+
+options:
+  project_src:
+    description:
+      - Path to a directory containing a C(docker-compose.yaml) file.
+      - Required.
+    type: path
+  state:
+    description:
+      - Desired state of the project.
+      - Specifying C(present) is the same as running C(docker-compose up) resp. C(docker-compose stop) (with I(stopped)) resp. C(docker-compose restart)
+        (with I(restarted)).
+      - Specifying C(absent) is the same as running C(docker-compose down).
+    type: str
+    default: present
+    choices:
+      - absent
+      - present
+  recreate:
+    description:
+      - By default containers will be recreated when their configuration differs from the service definition.
+      - Setting to C(never) ignores configuration differences and leaves existing containers unchanged.
+      - Setting to C(always) forces recreation of all existing containers.
+    type: str
+    default: smart
+    choices:
+      - always
+      - never
+      - smart
+  build:
+    description:
+      - Use with I(state) C(present) to always build images prior to starting the application.
+      - Same as running C(docker-compose build) with the pull option.
+      - Images will only be rebuilt if Docker detects a change in the Dockerfile or build directory contents.
+      - Use the I(nocache) option to ignore the image cache when performing the build.
+      - If an existing image is replaced, services using the image will be recreated unless I(recreate) is C(never).
+    type: bool
+    default: false
+  pull:
+    description:
+      - Use with I(state) C(present) to always pull images prior to starting the application.
+      - Same as running C(docker-compose pull).
+      - When a new image is pulled, services using the image will be recreated unless I(recreate) is C(never).
+    type: bool
+    default: false
+  nocache:
+    description:
+      - Use with the I(build) option to ignore the cache during the image build process.
+    type: bool
+    default: false
+  remove_images:
+    description:
+      - Use with I(state) C(absent) to remove all images or only local images.
+    type: str
+    choices:
+        - 'all'
+        - 'local'
+  remove_volumes:
+    description:
+      - Use with I(state) C(absent) to remove data volumes.
+    type: bool
+    default: false
+  remove_orphans:
+    description:
+      - Remove containers for services not defined in the Compose file when not C(restarted) or C(stopped).
+    type: bool
+    default: false
+  stopped:
+    description:
+      - Use with I(state) C(present) to stop all containers defined in the Compose file.
+    type: bool
+    default: false
+  restarted:
+    description:
+      - Use with I(state) C(present) to restart all containers defined in the Compose file.
+    type: bool
+    default: false
+
+requirements:
+  - "L(Python on Whales,https://pypi.org/project/python-on-whales/) >= 0.55.0"
+  - "docker-compose > 2.0.0"
+
+author:
+    - Sid Sun (@SidSun)
+'''
+
+EXAMPLES = r'''
+# Set up a virtual environment for the module and install python on whales
+- name: Set up virtual env for compose v2
+  ansible.builtin.pip:
+    name: 
+    - python-on-whales
+    virtualenv: ~/ansible-compose-v2
+
+# Ensure project directory is present
+- name: Create directory if it does not exist
+  ansible.builtin.file:
+    path: uptime-kuma
+    state: directory
+# Create the docker compose file 
+- name: Create compose file with content
+  copy:
+    dest: "uptime-kuma/docker-compose.yaml"
+    content: |
+      version: '3.3'
+
+      services:
+        uptime-kuma:
+          image: louislam/uptime-kuma:1
+          container_name: uptime-kuma
+          volumes:
+            - ./uptime-kuma-data:/app/data
+          ports:
+            - 3001:3001
+          restart: always
+
+# Tear down any containers, remove images and volumes
+- name: Tear down all existing containers
+  vars:
+    ansible_python_interpreter: ansible-compose-v2/bin/python
+  community.docker.docker_compose_v2:
+    project_src: uptime-kuma/docker-compose.yaml
+    remove_volumes: true
+    remove_orphans: true
+    remove_images: all
+    state: absent
+
+# Pull the latest image and then start services
+- name: Pull image and start service
+  vars:
+    ansible_python_interpreter: ansible-compose-v2/bin/python
+  community.docker.docker_compose_v2:
+    project_src: uptime-kuma/docker-compose.yaml
+    pull: true
+    state: present
+
+# Restart services
+- name: Restart service
+  vars:
+    ansible_python_interpreter: ansible-compose-v2/bin/python
+  community.docker.docker_compose_v2:
+    project_src: uptime-kuma/docker-compose.yaml
+    restarted: true
+    state: present
+
+# Stop services
+- name: Stop service
+  vars:
+    ansible_python_interpreter: ansible-compose-v2/bin/python
+  community.docker.docker_compose_v2:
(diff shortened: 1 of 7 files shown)


plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
...
+)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: docker_compose_v2
+
+short_description: Manage multi-container Docker applications with Docker Compose v2.
+
+version_added: "1.0.0"
```
felixfontein, line 18:
```suggestion
version_added: "3.4.0"
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
...
+
+description:
+  - Uses Docker Compose v2 to manage docker services. B(This module requires docker-compose > 2.0.0.)
```
felixfontein, line 21: Why does 2.0.0 not work? Or do you mean:
```suggestion
  - Uses Docker Compose v2 to manage docker services. B(This module requires docker-compose >= 2.0.0.)
```
?

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
...
+  - Configuration can be read from a C(docker-compose.yaml) file.
+  - See the examples for more details.
+  - Supports check mode.
+
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
```
felixfontein, line 30: You also need to add the correct docs fragment for this:
```
extends_documentation_fragment:
  - community.docker.attributes
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
+#!/usr/bin/python
+
+from __future__ import (absolute_import, division, print_function)
+from ansible.module_utils.basic import AnsibleModule
+import os
+from python_on_whales.docker_client import DockerClient
+from python_on_whales.exceptions import (
+    DockerException,
+)
+__metaclass__ = type
```
felixfontein, line 10: Please move everything that's not `metaclass` and `__future__` below the documentation (as in all other modules).

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
+#!/usr/bin/python
+
+from __future__ import (absolute_import, division, print_function)
+from ansible.module_utils.basic import AnsibleModule
+import os
+from python_on_whales.docker_client import DockerClient
+from python_on_whales.exceptions import (
```
felixfontein, line 7: Imports of third-party libraries need to be guarded (see https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/dev_guide/testing/sanity/import.rst#import).

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
+#!/usr/bin/python
+
```
felixfontein, line 2: You need to add copyright statements. Since you copied parts of the old module (documentation), you need to include its copyright statements as well next to yours.

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
...
+    default: false
+  restarted:
+    description:
+      - Use with I(state) C(present) to restart all containers defined in the Compose file.
+    type: bool
+    default: false
+
+requirements:
+  - "L(Python on Whales,https://pypi.org/project/python-on-whales/) >= 0.55.0"
+  - "docker-compose > 2.0.0"
```
felixfontein, line 111:
```suggestion
  - "docker-compose >= 2.0.0"
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
...
+    virtualenv: ~/ansible-compose-v2
+
+# Ensure project directory is present
+- name: Create directory if it does not exist
+  ansible.builtin.file:
+    path: uptime-kuma
+    state: directory
+# Create the docker compose file 
+- name: Create compose file with content
+  copy:
```
felixfontein, line 132:
```suggestion
  ansible.builtin.copy:
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
...
+    state: present
+'''
+
+RETURN = r'''
+# retuens are a bit finicky at the moment, see: https://github.com/gabrieldemarmiesse/python-on-whales/issues/395
+msg:
+    description: In docker failures, has docker command + some description. In other failures, stringified exception is returned. In success, descriptive message of what happened.
+    type: bool
+    returned: always
+    sample: 'true'
```
felixfontein, line 201:
```suggestion
    sample: true
```
Same below.

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
...
+  vars:
+    ansible_python_interpreter: ansible-compose-v2/bin/python
+  community.docker.docker_compose_v2:
+    project_src: uptime-kuma/docker-compose.yaml
+    recreate: always
+    state: present
+'''
+
+RETURN = r'''
+# retuens are a bit finicky at the moment, see: https://github.com/gabrieldemarmiesse/python-on-whales/issues/395
```
felixfontein, line 196:
```suggestion
# returns are a bit finicky at the moment, see: https://github.com/gabrieldemarmiesse/python-on-whales/issues/395
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
...
+    description: List stating which compose commands were run (not the entire command itself).
+    type: list
+    returned: success
+    sample: ["pull", "up"]
+rc:
+    description: Return code of executed docker command.
+    type: str
+    returned: failure from docker
+    sample: 12
+stderr_lines:
```
felixfontein, line 217: Common return values should not be documented (https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#common-return-values).

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
...
+        failure = dict(
+            msg=f"something went wrong, stderr, stdout and return code is printed, command executed: {e.docker_command}",
+            changed=True,
+            failed=True,
+            stderr=e.stderr,
+            stdout=e.stdout,
+            stderr_lines=e.stderr.split(
+                "\n") if e.stderr is not None else None,
+            stdout_lines=e.stdout.split(
+                "\n") if e.stdout is not None else None,
```
felixfontein, line 361:
```suggestion
```
Ansible will add these automatically when `stdout` / `stderr` are present.

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
...
+            rc=e.return_code,
+            err=str(e)
```
felixfontein, line 363:
```suggestion
            err=str(e),
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,371 @@
...
+    type: list
+    returned: failure from docker
+    sample: ["Authorization to docker daemon failed"]
+'''
+
+
+class ComposeV2Module:
+    module: AnsibleModule
+
+    def __init__(self, module: AnsibleModule) -> None:
```
felixfontein, line 233: Please remove all type hints. The code needs to be parsable by Python 2.x as well.
```suggestion
    def __init__(self, module: AnsibleModule):
```

felixfontein (commented):
Thanks for your contribution! I've added some first comments. Please also observe the failing tests.

felixfontein:
@Sid-Sun I pushed some basic tests for the module to your branch. I've added the setup role in another PR, so if you want to run the tests locally, you first have to rebase your branch against the current `main` branch.

To run the tests directly, you can add a playbook `tests/integration/targets/docker_compose_v2/run-tests-locally.yml` with the following content:
```
- hosts: localhost
  tasks:
  - name: Check Docker API version
    command: "{{ ansible_python.executable }} -c 'import docker; print(docker.from_env().version()[\"ApiVersion\"])'"
    register: docker_api_version_stdout
    ignore_errors: yes

  - set_fact:
      docker_api_version: "{{ docker_api_version_stdout.stdout or '0.0' }}"
      remote_tmp_dir: "{{ 'output-%0x' % ((2**32) | random) }}"
      has_docker_compose: true

  - debug:
      msg: "Docker API version: {{ docker_api_version }}"

  - include_vars: ../setup_docker/vars/main.yml

  - block:
    - name: Create output directory
      file:
        path: "{{ remote_tmp_dir }}"
        state: directory
    - include_tasks: tasks/main.yml
      vars:
        role_path: '.'
    always:
    - name: Remove output directory
      file:
        path: "{{ remote_tmp_dir }}"
        state: absent
```
and run `ansible-playbook run-tests-locally.yml` in that directory. Note that the playbook uses your local Docker Daemon; it tries to clean up, but if the module is broken you might have some leftovers. (Right now cleanup works.)

(When you run the tests you will notice that some of the tests are currently failing, for example check mode and idempotency.)

gotmax23:
python-on-whales only supports Python 3.7 and above. Shouldn't that be mentioned in the module docs and the README?

jkossis:
I know this is still a WIP, but I was curious if the plan is for this implementation to reflect the idempotent behavior of v1.

For example, in v1, [pulling an image is only marked as changed when the image is in fact new/different](https://github.com/ansible-collections/community.docker/blob/main/plugins/modules/docker_compose.py#L973). Whereas here, [it's marked as changed as long as we aren't running under `check_mode`](https://github.com/ansible-collections/community.docker/blob/6601a395796146fd55e3bcf92d26e6f9b4ab13cd/plugins/modules/docker_compose_v2.py#L290). This could be extended to other operations, just used pull as an example.

Sid-Sun:
thanks for the tests @felixfontein idempotency requires getting output from python-on-wheels and the current logic in POW doesn't allow for returning output, i'll try reworking it

felixfontein:
Idempotency is a basic feature that almost all Ansible modules have (and should have). The exceptions are mainly cases where idempotency is impossible, for example the `ansible.builtin.command` module, since that module cannot know whether a random command will result in a change or not, or APIs that do not allow to query certain information (for security reasons, because it is computationally infeasible, ...), etc. I think that for most cases it should be possible to implement idempotency for this module (pulling images is one of the cases where it is not so easy to implement, and probably impossible in case of check mode).

lel-amri:
Hello there,

I just pulled code today and noticed the setup_docker_compose_v2 test files, indicating that work was ongoing for supporting docker-compose v2.

I am currently working on this as well, although I took the `module.run_command()` approach as "hinted" in [this comment](https://github.com/ansible-collections/community.docker/issues/364#issuecomment-1201453058). I took inspiration in how POW handles docker-compose and how the `containers.podman` collection handles `podman kube play`.

I believe we should let docker-compose manage indempotency all by itself unless we want to handle _potentially_ clashing behavior. Getting the result of a docker-composer run is definitely achievable by reading what is written on stdout (while using the `--ansi never` option), but we have to keep in mind that output stability is not guaranteed. Check mode, though, seems impossible.

Now, about my ongoing work, I just want to get the job done and I don't see the point of both of us working on different implementations. I'm uneasy because I'd like to join the effort of supporting docker-compose v2, but after working on this I believe that the `module.run_command()` way is better than the POW way for two reasons : there is less dependencies and POW does not returns the output nor parses what docker-compose spits out. What do you think ?

Sid-Sun:
I agree,we should let docker compose manage idempotency. 

POW for this usecase seems to be a bit unideal, I have an open PR to change the API of POW to at least return the output though it hasn't gone anywhere yet. 

Thanks for working on this! hoping to see your implementation merge!😀

Sid-Sun:
I'll close this PR for now.

felixfontein:
@Sid-Sun thanks for working on this!

PR #545, opened 2023-01-02
Add docker_container_copy_into module

felixfontein:
##### SUMMARY
Create a module to copy files into a container.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
docker_container_copy_into


plugins/module_utils/copy.py
```diff
@@ -0,0 +1,339 @@
...
+                lambda client: client.get_raw_stream(
+                    '/containers/{0}/archive', container,
+                    params={'path': in_path},
+                    headers={'Accept-Encoding': 'identity'},
+                )
+            )
+        except DockerFileNotFound:
+            return process_none(in_path)
+
+        # TODO: stream tar file instead of downloading it into a BytesIO
```
briantist, line 170: I see the note in the doc about not using this for large files; it would be nice to have it use streams so arbitrarily large files could be supported.

plugins/modules/docker_container_copy_into.py
```diff
@@ -0,0 +1,327 @@
...
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+
+options:
+  container:
+    description:
+      - The name of the container to copy files to/from.
```
briantist, line 32:
```suggestion
      - The name of the container to copy files to.
```

Current module does not seem to support copying files out of the container.

briantist (commented):
I have two suggestions (that could be done in other PRs):

A `content` option (a la `copy` module) for writing small-ish content directly without having an actual source file.

A `base64_content` (name could be changed) that does similar to `content` but you supply it base64 pre-encoded content, which is decoded on the target.

plugins/modules/docker_container_copy_into.py
felixfontein: Thanks for spotting this! I originally wanted the module to support both directions, but then decided to better split that up into two modules (and start with the more interesting direction).

plugins/module_utils/copy.py
felixfontein: I've been playing around with streaming a bit, I think the easiest way is to not use Python's `tarfile`, but instead use some own code that allows to stream the contents of the generated tarfile. I'll play around with it some more...

felixfontein: I think this should be doable without too much repetition; `tarfile.TarInfo` does part of the heavy lifting (i.e. composing the file header) and can be used separately.

felixfontein: Ok, I've implemented that in 052c7f6214a619768ffe289fe106246918cc5eff.

felixfontein:
Things left to implement:
- [ ] diff mode
- [ ] `force=true` (disable idempotency checks)
- [ ] `content` and `content_base64`, or `content` and `content_is_b64` options (the latter is used by docker_config and docker_secret)

felixfontein:
Ok, I think this is ready for review again :)

plugins/modules/docker_container_copy_into.py
```diff
@@ -262,29 +596,194 @@ def copy_into_container(client, container, managed_path, container_path, follow_
...
         local_follow=dict(type='bool', default=True),
         owner_id=dict(type='int'),
         group_id=dict(type='int'),
         mode=dict(type='int'),
+        force=dict(type='bool'),
+        content=dict(type='str', no_log=True),
+        content_is_b64=dict(type='bool', default=False),
+
+        # Undocumented parameters for use by the action plugin
+        _max_file_size_for_diff=dict(type='int'),
```
briantist, line 775: Should this be exposed as a documented/configurable option for the module, which defaults to the Ansible configuration value if not otherwise set? Or do you think it's better not to let this be set independently?

briantist (approved):
None of my inline questions/comments are blocking, this is great stuff!

plugins/module_utils/copy.py
```diff
@@ -157,39 +298,29 @@ def fetch_file_ex(call_client, container, in_path, process_none, process_regular
...
-                    params={'path': in_path},
-                    headers={'Accept-Encoding': 'identity'},
-                )
+            stream = client.get_raw_stream(
+                '/containers/{0}/archive', container,
+                params={'path': in_path},
+                headers={'Accept-Encoding': 'identity'},
             )
-        except DockerFileNotFound:
+        except NotFound as dummy:
```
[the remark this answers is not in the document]
felixfontein:
```suggestion
        except NotFound:
```

plugins/modules/docker_container_copy_into.py
felixfontein: Hmm, I think for now I'd like to keep it this way. It can still be renamed / documented at any point.

felixfontein:

Issue #435, opened 2022-07-31
Differences between vendored Docker SDK for Python code and Docker SDK for Python


felixfontein:
##### SUMMARY
This issue is for tracking differences that need to be checked (whether our vendored code needs to be updated accordingly) that appear in the Docker SDK for Python in https://github.com/docker/docker-py/commits/main

##### ISSUE TYPE
- Tracking

##### COMPONENT NAME
plugins/module_utils/_api/

felixfontein:
#434 updates to up to https://github.com/docker/docker-py/commit/3f0095a7c1966c521652314e524ff362c24ff58c. Not yet there: https://github.com/docker/docker-py/compare/3f0095a7c1966c521652314e524ff362c24ff58c..main

felixfontein:
#438 updates to up to https://github.com/docker/docker-py/commit/42789818bed5d86b487a030e2e60b02bf0cfa284. Not yet there: https://github.com/docker/docker-py/compare/42789818bed5d86b487a030e2e60b02bf0cfa284..main

felixfontein:
No update required for current state https://github.com/docker/docker-py/commit/e901eac7a8c5f29c7720eafb9f58c8356cca2324. Not yet there: https://github.com/docker/docker-py/compare/e901eac7a8c5f29c7720eafb9f58c8356cca2324..main

felixfontein:
Docker SDK for Python 6.0.0 has been released. All changes in it have been incorporated (i.e. in the last 10 days there have been no new commits which affect code that's vendored here).

felixfontein:
No update required for current state https://github.com/docker/docker-py/commit/923e067dddc3d4b86e4e620a99fcdcdafbd17a98. Not yet there: https://github.com/docker/docker-py/compare/923e067dddc3d4b86e4e620a99fcdcdafbd17a98..main

felixfontein:
Not yet there: https://github.com/docker/docker-py/compare/8590eaad3c4b1460606763332ab84b70033ad6a1..main

felixfontein:
#581 and #582 update this to https://github.com/docker/docker-py/compare/aca129dd69fa068da247d37266110c3a9889a780...main

felixfontein:
#619 updates this to https://github.com/docker/docker-py/compare/9cadad009e6aa78e15d752e2011705d7d91b8d2e...main.

felixfontein:
[...]

PR #581, opened 2023-02-11
docker-py fix: fix(store): warn on init instead of throw

felixfontein:
##### SUMMARY
Cherry-picked https://github.com/docker/docker-py/pull/3080. 

Using the proper ansible-core mechanisms for warnings instead of `import warnings` resulted in a lot of extra code.

Ref: #435

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
vendored Docker SDK for Python

### DIFF
--- a/changelogs/fragments/581-store-warn-instead-of-error.yml
+++ b/changelogs/fragments/581-store-warn-instead-of-error.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - modules and plugins communicating directly with the Docker daemon - fail when using a credential helper instead of when initializing it (https://github.com/ansible-collections/community.docker/pull/581).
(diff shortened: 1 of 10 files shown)


kristianheljas:
I have a few issues with this one:

1. `warn` argument is ambiguous - I was surprised when i discovered that it's a function, not a boolean.
2. Mere presence of the warning function will [suppress an exception in `Store.__init()`](https://github.com/ansible-collections/community.docker/pull/581/files#diff-6048483085e39a11b53ecf243feea9242b3743d06aaa94ef13132c65fec9e9cfR36-R37).
3. It's gonna fail with the same message in [`Store._execute()` ](https://github.com/ansible-collections/community.docker/pull/581/files#diff-6048483085e39a11b53ecf243feea9242b3743d06aaa94ef13132c65fec9e9cfR87-R90) anyway so I can't see any advantage for the warning.

What are your thoughts about just silencing the [InitializationError](https://github.com/ansible-collections/community.docker/pull/581/files#diff-6048483085e39a11b53ecf243feea9242b3743d06aaa94ef13132c65fec9e9cfR39) via explicit argument - `validate_program_path=False` for example?

felixfontein:
Let's close this, as it isn't really needed anyway.

PR #586, opened 2023-02-12
Support docker compose v2

### LINKED ISSUE #216: docker-compose 2.0.0 is no longer written in Python, thus breaking the docker_compose module

lel-amri:
##### SUMMARY
Implementation of a module that supports docker-compose version 2.

The module has no dependency on external libraries. It relies on `AnsibleModule.run_command()` to execute `docker-compose` and on `ansible.module_utils.common.yaml.yaml_dump()` for dumping `definition`'s content to `docker-compose`'s stdin.

This module answers 90% of my needs. I tried to make it as flexible as possible, thus it is not as DRY as it could be. I'm not sure what path to take regarding idempotency in some case. I'd prefer to get feedback from the community about the various the usages of the `docker_compose` module and needs for the `docker_compose_v2` module.

Should fix #216.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
docker_compose_v2

##### ADDITIONAL INFORMATION
I mostly took decisions based on what was discussed in #364.
Finally, this PR is mostly here to take advantage of the CI (and to allow comparison with @Sid-Sun's work at #540).

### DIFF
--- a/plugins/module_utils/_yaml.py
+++ b/plugins/module_utils/_yaml.py
@@ -1,12 +1,8 @@
-# PLEASE READ THE FOLLOWING
-# This file is the vendored version of Ansible's `/lib/ansible/module_utils/common/yaml.py`.
-# This file MUST be removed once this collection drops support for ansible-core v2.11, and
-# related imports MUST be changed to `from ansible.module_utils.common.yaml import ...` or
-# alike.
-# This file was taken at https://raw.githubusercontent.com/ansible/ansible/5f6ebebce9f0e7eab1ba26848d2dd02d24f43e0a/lib/ansible/module_utils/common/yaml.py
-
-# (c) 2020 Matt Martz <matt@sivel.net>
-# Simplified BSD License (see licenses/simplified_bsd.txt or https://opensource.org/licenses/BSD-2-Clause)
+# A vendored version of ansible-core's `lib/ansible/module_utils/common/yaml.py`.
+# This file can be removed once community.docker drops support for ansible-core 2.11.
+# Copyright (c) 2020 Matt Martz <matt@sivel.net>
+# Simplified BSD License (see LICENSES/BSD-2-Clause.txt or https://opensource.org/licenses/BSD-2-Clause)
+# SPDX-License-Identifier: BSD-2-Clause
 
 """
 This file provides ease of use shortcuts for loading and dumping YAML,
--- a/plugins/module_utils/_yaml.py
+++ b/plugins/module_utils/_yaml.py
@@ -1,0 +1,55 @@
+# PLEASE READ THE FOLLOWING
+# This file is the vendored version of Ansible's `/lib/ansible/module_utils/common/yaml.py`.
+# This file MUST be removed once this collection drops support for ansible-core v2.11, and
+# related imports MUST be changed to `from ansible.module_utils.common.yaml import ...` or
+# alike.
+# This file was taken at https://raw.githubusercontent.com/ansible/ansible/5f6ebebce9f0e7eab1ba26848d2dd02d24f43e0a/lib/ansible/module_utils/common/yaml.py
+
+# (c) 2020 Matt Martz <matt@sivel.net>
+# Simplified BSD License (see licenses/simplified_bsd.txt or https://opensource.org/licenses/BSD-2-Clause)
+
+"""
+This file provides ease of use shortcuts for loading and dumping YAML,
+preferring the YAML compiled C extensions to reduce duplicated code.
+"""
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from functools import partial as _partial
+
+HAS_LIBYAML = False
+
+try:
+    import yaml as _yaml
+except ImportError:
+    HAS_YAML = False
+else:
+    HAS_YAML = True
+
+if HAS_YAML:
+    try:
+        from yaml import CSafeLoader as SafeLoader
+        from yaml import CSafeDumper as SafeDumper
+        from yaml.cyaml import CParser as Parser  # type: ignore[attr-defined]  # pylint: disable=unused-import
+
+        HAS_LIBYAML = True
+    except (ImportError, AttributeError):
+        from yaml import SafeLoader  # type: ignore[assignment]
+        from yaml import SafeDumper  # type: ignore[assignment]
+        from yaml.parser import Parser  # type: ignore[assignment]  # pylint: disable=unused-import
+
+    yaml_load = _partial(_yaml.load, Loader=SafeLoader)
+    yaml_load_all = _partial(_yaml.load_all, Loader=SafeLoader)
+
+    yaml_dump = _partial(_yaml.dump, Dumper=SafeDumper)
+    yaml_dump_all = _partial(_yaml.dump_all, Dumper=SafeDumper)
+else:
+    SafeLoader = object  # type: ignore[assignment,misc]
+    SafeDumper = object  # type: ignore[assignment,misc]
+    Parser = object  # type: ignore[assignment,misc]
+
+    yaml_load = None  # type: ignore[assignment]
+    yaml_load_all = None  # type: ignore[assignment]
+    yaml_dump = None  # type: ignore[assignment]
+    yaml_dump_all = None  # type: ignore[assignment]
--- a/plugins/modules/docker_compose_v2.py
+++ b/plugins/modules/docker_compose_v2.py
@@ -1,0 +1,871 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+
+DOCUMENTATION = '''
+
+module: docker_compose_v2
+
+short_description: Manage multi-container Docker applications with Docker Compose.
+
+author: Léo El Amri (@lel-amri)
+
+description:
+  - Uses Docker Compose to start and shutdown services.
+  - Swarm mode is not supported (thus secrets and configs are not supported).
+  - Configuration can be read from a Compose file or inline using the I(definition) option.
+  - See the examples for more details.
+
+attributes:
+  check_mode:
+    support: none
+  diff_mode:
+    support: none
+
+options:
+  docker_host:
+    description:
+      - The URL or Unix socket path used to connect to the Docker API. To connect to a remote host, provide the
+        TCP connection string. For example, C(tcp://192.0.2.23:2376). If TLS is used to encrypt the connection,
+        the module will automatically replace C(tcp) in the connection URL with C(https).
+      - If the value is not specified in the task, the value of environment variable C(DOCKER_HOST) will be used
+        instead. If the environment variable is not set, the default value will be used.
+    type: str
+    default: unix://var/run/docker.sock
+    aliases: [ docker_url ]
+  project_name:
+    description:
+      - Provide a project name.
+      - Equivalent to C(docker-compose --project-name).
+    type: str
+  env_file:
+    description:
+      - By default environment files are loaded from a C(.env) file located directly under the I(project_src) directory.
+      - I(env_file) can be used to specify the path of a custom environment file instead.
+      - The path is relative to the I(project_src) directory.
+      - Equivalent to C(docker-compose --env-file).
+    type: path
+  files:
+    description:
+      - List of Compose files.
+      - Files are passed to docker-compose in the order given.
+      - Equivalent to C(docker-compose -f).
+    type: list
+    elements: path
+  profiles:
+    description:
+      - List of profiles to enable when starting services.
+      - Equivalent to C(docker-compose --profile).
+    type: list
+    elements: str
+  state:
+    description:
+      - Desired state of the project.
+      - Specifying C(pulled) is the same as running C(docker-compose pull).
+      - Specifying C(built) is the same as running C(docker-compose build).
+      - Specifying C(stopped) is the same as running C(docker-compose stop).
+      - Specifying C(present) is the same as running C(docker-compose up).
+      - Specifying C(restarted) is the same as running C(docker-compose restart).
+      - Specifying C(absent) is the same as running C(docker-compose down).
+    type: str
+    default: present
+    choices:
+      - absent
+      - built
+      - present
+      - pulled
+      - restarted
+      - stopped
+  services:
+    description:
+      - When I(state) is C(present) run C(docker-compose up) resp. C(docker-compose stop) (with I(stopped)) resp. C(docker-compose restart) (with I(restarted))
+        on a subset of services.
+      - If empty, which is the default, the operation will be performed on all services defined in the Compose file (or inline I(definition)).
+    type: list
+    elements: str
+  dependencies:
+    description:
+      - When I(state) is C(present) specify whether or not to include linked services.
+      - When false, equivalent to C(docker-compose up --no-deps).
+      - When I(state) is C(pull) specify whether or not to also pull for linked services.
+      - When true, equivalent to C(docker-compose pull --include-deps).
+    type: bool
+    default: true
+  definition:
+    description:
+      - Compose file describing one or more services, networks and volumes.
+      - Mutually exclusive with I(files).
+    type: dict
+  recreate:
+    description:
+      - By default containers will be recreated when their configuration differs from the service definition.
+      - Setting to C(never) ignores configuration differences and leaves existing containers unchanged.
+      - Setting to C(always) forces recreation of all existing containers.
+      - When set to C(never), equivalent to C(docker-compose up --no-recreate).
+      - When set to C(always), equivalent to C(docker-compose up --force-recreate).
+    type: str
+    default: smart
+    choices:
+      - always
+      - never
+      - smart
+  build:
+    description:
+      - Use with I(state) C(present) to always build images prior to starting the application.
+      - Equivalent to C(docker-compose up --build).
+      - Images will only be rebuilt if Docker detects a change in the Dockerfile or build directory contents.
+      - If an existing image is replaced, services using the image will be recreated unless I(recreate) is C(never).
+    type: bool
+    default: false
+  pull:
+    description:
+      - Use with I(state) C(present) to always pull images prior to starting the application.
+      - Equivalent to C(docker-compose up --pull always).
+      - When a new image is pulled, services using the image will be recreated unless I(recreate) is C(never).
+    type: bool
+    default: false
+  nocache:
+    description:
+      - Use with the I(build) option to ignore the cache during the image build process.
+      - Equivalent to C(docker-compose build --no-cache).
+    type: bool
+    default: false
+  remove_images:
+    description:
+      - Use with I(state) C(absent) to remove all images or only local images.
+      - Equivalent to C(docker-compose down --rmi all|local).
+    type: str
+    choices:
+        - 'all'
+        - 'local'
+  remove_volumes:
+    description:
+      - Use with I(state) C(absent) to remove data volumes.
+      - Equivalent to C(docker-compose down --volumes).
+    type: bool
+    default: false
+  remove_orphans:
+    description:
+      - Remove containers for services not defined in the Compose file.
+      - Equivalent to C(docker-compose up --remove-orphans) or C(docker-compose down --remove-orphans).
+    type: bool
+    default: false
+  timeout:
+    description:
+      - Timeout in seconds for container shutdown when attached or when containers are already running.
+      - By default C(compose) will use a C(10s) timeout unless C(default_grace_period) is defined for a
+        particular service in the I(project_src).
+    type: int
+    default: null
+
+requirements:
+  - "docker-compose >= 2.0.0"
+'''
+
+EXAMPLES = '''
+# Examples use the django example at https://docs.docker.com/compose/django. Follow it to create the
+# flask directory
+
+- name: Run using a project directory
+  hosts: localhost
+  gather_facts: false
+  tasks:
+    - name: Tear down existing services
+      community.docker.docker_compose:
+        project_src: flask
+        state: absent
+
+    - name: Create and start services
+      community.docker.docker_compose:
+        project_src: flask
+      register: output
+
+    - ansible.builtin.debug:
+        var: output
+
+    - name: Run `docker-compose up` again
+      community.docker.docker_compose:
+        project_src: flask
+        build: false
+      register: output
+
+    - ansible.builtin.debug:
+        var: output
+
+    - ansible.builtin.assert:
+        that: not output.changed
+
+    - name: Stop all services
+      community.docker.docker_compose:
+        project_src: flask
+        build: false
+        stopped: true
+      register: output
+
+    - ansible.builtin.debug:
+        var: output
+
+    - ansible.builtin.assert:
+        that:
+          - "'stopped' in containers['flask_web_1'] | default([])"
+          - "'stopped' in containers['flask_db_1'] | default([])"
+
+    - name: Restart services
+      community.docker.docker_compose:
+        project_src: flask
+        build: false
+        restarted: true
+      register: output
+
+    - ansible.builtin.debug:
+        var: output
+
+    - ansible.builtin.assert:
+        that:
+          - "'started' in containers['flask_web_1'] | default([])"
+          - "'started' in containers['flask_db_1'] | default([])"
+
+- name: Run with inline Compose file version 2
+  # https://docs.docker.com/compose/compose-file/compose-file-v2/
+  hosts: localhost
+  gather_facts: false
+  tasks:
+    - community.docker.docker_compose:
+        project_src: flask
+        state: absent
+
+    - community.docker.docker_compose:
+        project_name: flask
+        definition:
+          version: '2'
+          services:
+            db:
+              image: postgres
+            web:
+              build: "{{ playbook_dir }}/flask"
+              command: "python manage.py runserver 0.0.0.0:8000"
+              volumes:
+                - "{{ playbook_dir }}/flask:/code"
+              ports:
+                - "8000:8000"
+              depends_on:
+                - db
+      register: output
+
+    - ansible.builtin.debug:
+        var: output
+
+    - ansible.builtin.assert:
+        that:
+          - "'started' in containers['flask_web_1'] | default([])"
+          - "'started' in containers['flask_db_1'] | default([])"
+
+- name: Run with inline Compose file version 1
+  # https://docs.docker.com/compose/compose-file/compose-file-v1/
+  hosts: localhost
+  gather_facts: false
+  tasks:
+    - community.docker.docker_compose:
+        project_src: flask
+        state: absent
+
+    - community.docker.docker_compose:
+        project_name: flask
+        definition:
+            db:
+              image: postgres
+            web:
+              build: "{{ playbook_dir }}/flask"
+              command: "python manage.py runserver 0.0.0.0:8000"
+              volumes:
+                - "{{ playbook_dir }}/flask:/code"
+              ports:
+                - "8000:8000"
+              links:
+                - db
+      register: output
+
+    - ansible.builtin.debug:
+        var: output
+
+    - ansible.builtin.assert:
+        that:
+          - "'started' in containers['flask_web_1'] | default([])"
+          - "'started' in containers['flask_db_1'] | default([])"
+'''
+
+RETURN = '''
+stdout:
+  description:
+  - The stdout from docker-compose.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: str
+stderr:
+  description:
+  - The stderr from docker-compose.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: str
+containers:
+  description:
+  - A dictionary mapping the various status of containers during C(docker-compose) operation.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: complex:
+  contains:
+    container_name:
+      description: Name of the container.
+      returned: always, unless when C(docker-compose) was not given the chance to run
+      type: list
+      elements: str
+      example: ["stopped", "removed"]
+volumes:
+  description:
+  - A dictionary mapping the various status of volumes during C(docker-compose) operation.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: complex:
+  contains:
+    volume_name:
+      description: Name of the volume.
+      returned: always, unless when C(docker-compose) was not given the chance to run
+      type: list
+      elements: str
+      example: ["created"]
+images:
+  description:
+  - A dictionary mapping the various status of volumes during C(docker-compose) operation.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: complex:
+  contains:
+    image_name:
+      description: Name of the image.
+      returned: always, unless when C(docker-compose) was not given the chance to run
+      type: list
+      elements: str
+      example: ["removed"]
+networks:
+  description:
+  - A dictionary mapping the various status of networks during C(docker-compose) operation.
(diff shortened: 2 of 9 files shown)


plugins/modules/docker_compose2.py
```diff
@@ -0,0 +1,866 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+
+DOCUMENTATION = '''
+
+module: docker_compose2
```
felixfontein, line 7:
```suggestion
module: docker_compose_v2

version_added: 3.5.0
```

plugins/modules/docker_compose2.py
```diff
@@ -0,0 +1,866 @@
...
+
+short_description: Manage multi-container Docker applications with Docker Compose.
+
+author: Léo El Amri (@lel-amri)
+
+description:
+  - Uses Docker Compose to start and shutdown services.
+  - Swarm mode is not supported (thus secrets and configs aren't supported).
```
felixfontein, line 15:
```suggestion
  - Swarm mode is not supported (thus secrets and configs are not supported).
```
(appears some more times in the return value docs)

plugins/modules/docker_compose2.py
```diff
@@ -0,0 +1,866 @@
...
+      type: list
+      elements: str
+      example: ["created"]
+'''
+
+
+from typing import List, Optional, Tuple, Union, Literal, Final, FrozenSet
+import re
+from collections import defaultdict
+import enum
```
felixfontein, line 362: Please note that `enum` is not available in Python 2.7.

plugins/modules/docker_compose2.py
```diff
@@ -0,0 +1,866 @@
...
+from dataclasses import dataclass
```
felixfontein, line 363: dataclasses is Python 3.7+ and isn't supported on Python 2.7, 3.5, and 3.6.

plugins/modules/docker_compose2.py
```diff
@@ -0,0 +1,866 @@
...
+            ('files', 'definition'),
+        ],
+    )
+    if bool(module.params['stopped']) and bool(module.params['restarted']):
+        module.fail_json(changed=False, msg='Cannot use restarted and stopped at the same time.')
+    changed = False
+    compose = ComposeManager(module, module.params['docker_host'])
+    if module.params['definition'] is not None:
+        if not HAS_YAML:
+            raise Exception('Require YAML support')
```
felixfontein, line 765: Please use `module.fail_json(msg=missing_required_lib('PyYAML'))`, where `from ansible.module_utils.basic import missing_required_lib`.

felixfontein (commented):
Thanks for your contribution! Please note that there is something strange going on in your branch, in https://github.com/ansible-collections/community.docker/pull/586/files you can see the module and a rename. I've added some more quick comments below.

> Finally, this PR is mostly here to take advantage of the CI

Please note that you need to add integration tests (similar as in #540) in tests/integration/targets/docker_compose_v2/ to fully make use of CI :)

plugins/modules/docker_compose2.py
lel-amri: Damn, I overlooked that this collection still supported Python 2.7. I knew I read it somewhere in #540. I'll also have to change type hint to the comment style and remove all occurrence of `/` and `*` parameters.

plugins/modules/docker_compose2.py
lel-amri: Replaced :)

plugins/modules/docker_compose2.py
lel-amri: I think I got them all.

plugins/modules/docker_compose2.py
lel-amri: Replaced with a crappy enum-like class (just a namespace of `object` instances).

plugins/modules/docker_compose2.py
lel-amri: Replaced with a `namedtuple`.

plugins/modules/docker_compose2.py
lel-amri: Nice catch. I've replaced this code with :
```suggestion
            module.fail_json(
                changed=False,
                msg=missing_required_lib('PyYAML', reason='when using the "definition" argument.')
            )
```

lel-amri:
> Thanks for your contribution! Please note that there is something strange going on in your branch, in https://github.com/ansible-collections/community.docker/pull/586/files you can see the module and a rename. I've added some more quick comments below.

I typo'ed when doing a `git mv`. I actually moved `docker_container.py` instead of `docker_container2.py`. This is fixed in https://github.com/ansible-collections/community.docker/pull/586/commits/c3b7ccc4cd39d37f5d5c842ff8736b22a0d500b0. I'm sorry that the history does not make any sense, the code I'm pushing isn't 1:1 the branch I'm working on. Nothing a squash can't fix :p

> > Finally, this PR is mostly here to take advantage of the CI
> 
> Please note that you need to add integration tests (similar as in #540) in tests/integration/targets/docker_compose_v2/ to fully make use of CI :)

I will !

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,926 @@
...
+  timeout:
+    description:
+      - Timeout in seconds for container shutdown when attached or when containers are already running.
+      - By default C(compose) will use a C(10s) timeout unless C(default_grace_period) is defined for a
+        particular service in the I(project_src).
+    type: int
+    default: null
+
+requirements:
+  - "docker-compose >= 2.0.0"
```
felixfontein, line 173:
```suggestion
  - "docker-compose >= 2.0.0"
  - PyYAML
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,926 @@
...
+        Text = str
+else:
+    TYPE_CHECKING = False
+import re
+from collections import defaultdict, namedtuple
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible_collections.community.docker.plugins.module_utils.util import (
+    DOCKER_COMMON_ARGS,
+)
+from ansible.module_utils.common.yaml import HAS_YAML, yaml_dump
```
felixfontein, line 397: Please note that this helper was only added in ansible-core 2.12. My suggestion would be to vendor https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/common/yaml.py as `plugins/module_utils/_yaml.py` in this collection, with a comment that it can be replaced by `ansible.module_utils.common.yaml` once support for ansible-core 2.11 has been dropped.

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,926 @@
...
+stderr:
+  description:
+    - The stderr from docker-compose.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: str
+containers:
+  description:
+    - A dictionary mapping the various status of containers during C(docker-compose) operation.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: complex
```
felixfontein, line 326:
```suggestion
  type: dict
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,926 @@
...
+      description: Name of the container.
+      returned: always, unless when C(docker-compose) was not given the chance to run
+      type: list
+      elements: str
+      example: ["stopped", "removed"]
+volumes:
+  description:
+    - A dictionary mapping the various status of volumes during C(docker-compose) operation.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: complex
```
felixfontein, line 338:
```suggestion
  type: dict
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,926 @@
...
+      description: Name of the volume.
+      returned: always, unless when C(docker-compose) was not given the chance to run
+      type: list
+      elements: str
+      example: ["created"]
+images:
+  description:
+    - A dictionary mapping the various status of volumes during C(docker-compose) operation.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: complex
```
felixfontein, line 350:
```suggestion
  type: dict
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,926 @@
...
+      description: Name of the image.
+      returned: always, unless when C(docker-compose) was not given the chance to run
+      type: list
+      elements: str
+      example: ["removed"]
+networks:
+  description:
+    - A dictionary mapping the various status of networks during C(docker-compose) operation.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: complex
```
felixfontein, line 362:
```suggestion
  type: dict
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,926 @@
...
+  description:
+    - A dictionary mapping the various status of volumes during C(docker-compose) operation.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: complex
+  contains:
+    volume_name:
+      description: Name of the volume.
+      returned: always, unless when C(docker-compose) was not given the chance to run
+      type: list
+      elements: str
```
felixfontein, line 344: If this is a single name, why is it a list of strings?

lel-amri: I think I misunderstand how dicts are built `RETURN`.

The elements `containers`, `images`, `volumes` and `networks` looks like this:
```json
{
  "name_of_the_resource": [
    "statuses",
    "it",
    "went",
    "through",
    "during",
    "run"
  ]
}
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,924 @@
...
+    - A dictionary mapping the various status of containers during C(docker-compose) operation.
+  returned: always, unless when C(docker-compose) was not given the chance to run
+  type: complex
+  contains:
+    container_name:
+      description: Name of the container.
+      returned: always, unless when C(docker-compose) was not given the chance to run
+      type: list
+      elements: str
+      example: ["stopped", "removed"]
```
felixfontein, line 333: If `containers` is a dictionary mapping strings (container names) to lists of strings (states of containers), then the above should be replaced by:
```suggestion
  sample:
    container_1:
      - stopped
      - removed
    container_2:
      - created
      - started
```

felixfontein: You are currently documenting a structure like
```
containers:
  container_name:
    - stopped
    - removed
```
where both `containers` and `container_name` are fixed strings and `containers` only ever contains one element (namely `container_name`).

lel-amri: Thanks a lot. I tried to make sense of this by reading Ansible's documentation, which is lacking on this matter IMO, then the source-code, which is very "bare-bones" regarding plugins documentation processing and doesn't explain much of what is _functionally_ expected.
From what you are saying, there is no way to document arbitrary keys for a dict with the `contains` structure ?

felixfontein: > From what you are saying, there is no way to document arbitrary keys for a dict with the `contains` structure ?

Yes, that's not possible.

lel-amri: Should be fixed. I retained the description.

plugins/modules/docker_compose_v2.py
lel-amri: Fixed, see https://github.com/ansible-collections/community.docker/pull/586#discussion_r1114211046

plugins/modules/docker_compose_v2.py
lel-amri: Fixed, see https://github.com/ansible-collections/community.docker/pull/586#discussion_r1114211046

plugins/modules/docker_compose_v2.py
lel-amri: Fixed, see https://github.com/ansible-collections/community.docker/pull/586#discussion_r1114211046

plugins/modules/docker_compose_v2.py
lel-amri: Fixed, see https://github.com/ansible-collections/community.docker/pull/586#discussion_r1114211046

plugins/modules/docker_compose_v2.py
lel-amri: Vendored

plugins/modules/docker_compose_v2.py
lel-amri: Done

tests/integration/targets/docker_compose_v2/tasks/tests/start-stop.yml
```diff
@@ -0,0 +1,241 @@
...
+
+    ####################################################################
+    ## Present #########################################################
+    ####################################################################
+
+    - name: Present (check)
+      docker_compose_v2:
+        state: present
+        project_src: "{{ remote_tmp_dir }}/{{ cname_prefix }}"
+      check_mode: true
```
felixfontein, line 53: You want to remove all check mode tests if the module does not support check mode.

plugins/module_utils/_yaml.py
```diff
@@ -0,0 +1,55 @@
+# PLEASE READ THE FOLLOWING
+# This file is the vendored version of Ansible's `/lib/ansible/module_utils/common/yaml.py`.
+# This file MUST be removed once this collection drops support for ansible-core v2.11, and
+# related imports MUST be changed to `from ansible.module_utils.common.yaml import ...` or
+# alike.
+# This file was taken at https://raw.githubusercontent.com/ansible/ansible/5f6ebebce9f0e7eab1ba26848d2dd02d24f43e0a/lib/ansible/module_utils/common/yaml.py
+
+# (c) 2020 Matt Martz <matt@sivel.net>
+# Simplified BSD License (see licenses/simplified_bsd.txt or https://opensource.org/licenses/BSD-2-Clause)
```
felixfontein, line 9:
```suggestion
# A vendored version of ansible-core's `lib/ansible/module_utils/common/yaml.py`.
# This file can be removed once community.docker drops support for ansible-core 2.11.

# Copyright (c) 2020 Matt Martz <matt@sivel.net>
# Simplified BSD License (see LICENSES/BSD-2-Clause.txt or https://opensource.org/licenses/BSD-2-Clause)
# SPDX-License-Identifier: BSD-2-Clause
```
You also need to add https://github.com/ansible/ansible/blob/devel/licenses/simplified_bsd.txt as `LICENSES/BSD-2-Clause.txt`, and add `BSD-2-Clause` to https://github.com/ansible-collections/community.docker/blob/main/galaxy.yml#L15-L17.

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
...
+description:
+  - Uses Docker Compose to start and shutdown services.
+  - Swarm mode is not supported (thus secrets and configs are not supported).
+  - Configuration can be read from a Compose file or inline using the I(definition) option.
+  - See the examples for more details.
+
+attributes:
+  check_mode:
+    support: none
+    description: Check mode is not supported because the `docker-compose` CLI doesn't allow it.
```
felixfontein, line 26:
```suggestion
    details: Check mode is not supported because the C(docker-compose) CLI does not allow it.
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
...
+  diff_mode:
+    support: none
+    description: Diff mode is not supported.
```
felixfontein, line 29:
```suggestion
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
...
+module: docker_compose_v2
+
+short_description: Manage multi-container Docker applications with Docker Compose.
+
+description:
+  - Uses Docker Compose to start and shutdown services.
+  - Swarm mode is not supported (thus secrets and configs are not supported).
+  - Configuration can be read from a Compose file or inline using the I(definition) option.
+  - See the examples for more details.
+
```
felixfontein, line 22:
```suggestion

extends_documentation_fragment:
  - community.docker.attributes

```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
...
+      - When set to C(always), equivalent to C(docker-compose up --force-recreate).
+    type: str
+    default: smart
+    choices:
+      - always
+      - never
+      - smart
+  build:
+    description:
+      - Use with I(state) C(present) to always build images prior to starting the application.
```
felixfontein, line 125:
```suggestion
      - Use with I(state=present) to always build images prior to starting the application.
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
...
+      - Equivalent to C(docker-compose up --build).
+      - Images will only be rebuilt if Docker detects a change in the Dockerfile or build directory contents.
+      - If an existing image is replaced, services using the image will be recreated unless I(recreate) is C(never).
+    type: bool
+    default: false
+  pull:
+    description:
+      - Use with I(state) C(present) to always pull images prior to starting the application.
```
felixfontein, line 133:
```suggestion
      - Use with I(state=present) to always pull images prior to starting the application.
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
...
+    default: false
+  nocache:
+    description:
+      - Use with the I(build) option to ignore the cache during the image build process.
+      - Equivalent to C(docker-compose build --no-cache).
+    type: bool
+    default: false
+  remove_images:
+    description:
+      - Use with I(state) C(absent) to remove all images or only local images.
```
felixfontein, line 146:
```suggestion
      - Use with I(state=absent) to remove all images or only local images.
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
...
+      - Equivalent to C(docker-compose down --rmi all|local).
+    type: str
+    choices:
+      - all
+      - local
+  remove_volumes:
+    description:
+      - Use with I(state) C(absent) to remove data volumes.
```
felixfontein, line 154:
```suggestion
      - Use with I(state=absent) to remove data volumes.
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
...
+  remove_orphans:
+    description:
+      - Remove containers for services not defined in the Compose file.
+      - Equivalent to C(docker-compose up --remove-orphans) or C(docker-compose down --remove-orphans).
+    type: bool
+    default: false
+  timeout:
+    description:
+      - Timeout in seconds for container shutdown when attached or when containers are already running.
+      - By default C(compose) will use a C(10s) timeout unless C(default_grace_period) is defined for a
```
felixfontein, line 167: Should `compose` be `docker-compose` here?
```suggestion
      - By default C(docker-compose) will use a C(10s) timeout unless C(default_grace_period) is defined for a
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
...
+        particular service in the I(project_src).
+    type: int
+    default: null
```
felixfontein, line 170:
```suggestion
```

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
...
+      - Compose file describing one or more services, networks and volumes.
+      - Mutually exclusive with I(files).
+    type: dict
+  recreate:
+    description:
+      - By default containers will be recreated when their configuration differs from the service definition.
+      - Setting to C(never) ignores configuration differences and leaves existing containers unchanged.
+      - Setting to C(always) forces recreation of all existing containers.
+      - When set to C(never), equivalent to C(docker-compose up --no-recreate).
+      - When set to C(always), equivalent to C(docker-compose up --force-recreate).
```
felixfontein, line 116: `smart` is not documented here.

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
...
+        the module will automatically replace C(tcp) in the connection URL with C(https).
+      - If the value is not specified in the task, the value of environment variable C(DOCKER_HOST) will be used
+        instead. If the environment variable is not set, the default value will be used.
+    type: str
+    default: unix://var/run/docker.sock
+    aliases: [ docker_url ]
+  project_src:
+    description:
+      - Path to the root directory of the Compose project.
+      - Required when I(definition) is provided.
```
felixfontein, line 45: I don't see an option called `definition`.

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,923 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright Léo El Amri
```
felixfontein, line 4: Please note that in case you copied significant parts of an existing module (like documentation), you need to add the copyright lines from that module here as well.

craph:
Hello, @felixfontein , @lel-amri , I don't understand how your code works with the new docker compose plugin written in Go.
Inside your Python code you are referencing `docker-compose` L.495 [docker_compose_v2.py](https://github.com/ansible-collections/community.docker/pull/586/files#diff-658ff6c2d6c0e440182b81fd9022a6247faf3730751f92c86fff5132842a9993)

how are you connecting with the API ?

Thank you very much for your help

lel-amri:
@craph 
> how are you connecting with the API ?

docker-compose is responsible for calling the docker daemon API. AFAIK there is no API offered by docker-compose. In previous versions of this module, we used the internals of docker-compose version 1, because it was implemented in Python. In version 2, since this is Go, there is no way for us to interface with the internals of docker-compose. Here I'm relying on the textual output of docker-compose to deduce what was done. There is a lot of limits with this approach, the biggest being that the textual output isn't guaranteed to be stable, thus this module could break easily any time in the future.
To implement this module I read docker-compose's source code, especially the `pkg/progress/plain.go` and `pkg/progress/event.go` files.

craph:
@lel-amri Thank you very much for the update.
But, if I install the latest version of docker-ce with the docker-ce-cli and docker-compose-plugin to have the latest version of docker compose V2.16.0 the command `docker-compose` doesn't exists anymore if I understand correctly the documentation.
The only thing available is to create an alias `docker-compose` that calls `docker compose` the new way to call compose V2.

I was just curious to understand how it is working to call compose v2

lel-amri:
@craph 
> if I install the latest version of docker-ce with the docker-ce-cli and docker-compose-plugin to have the latest version of docker compose V2.16.0 the command `docker-compose` doesn't exists anymore

Indeed, I overlooked that. I'll come up with a fix in this PR. This only works with the `docker-compose` standalone for now. In the meantime, the "alias" of `docker-compose` to `docker compose` will work as a workaround.

plugins/modules/docker_compose_v2.py
lel-amri: I believe the docker_compose_v2 module does not support what's documented in this fragment.

plugins/modules/docker_compose_v2.py
lel-amri: Indeed, it should be docker-compose.

plugins/modules/docker_compose_v2.py
lel-amri: There is one :
* https://github.com/ansible-collections/community.docker/blob/7c109a366a6a507f9d9aa82a95f24ca19a40c21b/plugins/modules/docker_compose_v2.py#L94
* https://github.com/ansible-collections/community.docker/blob/7c109a366a6a507f9d9aa82a95f24ca19a40c21b/plugins/modules/docker_compose_v2.py#L737

plugins/modules/docker_compose_v2.py
lel-amri: It is implied on the line right above (and because the default value for this choice is `smart`):

https://github.com/ansible-collections/community.docker/blob/5d05bedc250356dce166e285bcc6484773329b7e/plugins/modules/docker_compose_v2.py#L112

Should I expand as follow ?:

> By default (when set to C(smart)) containers will be recreated when their configuration differs from the service definition.

I'd prefer to remove the default value for the choice, but I'm not sure this is allowed.

craph:
Any update about the support of docker compose version 2 ?

lel-amri:
I have to finish writing tests. I've already implemented the fix for supporting docker compose as a docker plugin, but didn't pushed it yet.

Also, I have encountered an issue in a real-world scenario. Docker compose got an update that modifies what is written on stderr. That broke the module's reporting (not reporting changed when it should have). I think we will have to ask for stability on stderr, or a way to make this machine parseable. This would also benefit POW.

shantanoo-desai:
@lel-amri is it because of the ANSI graphics on the terminal? I think you can use `docker compose --ansi=never`

lel-amri:
> @lel-amri is it because of the ANSI graphics on the terminal? I think you can use `docker compose --ansi=never`

That's already what I'm doing ;) :

https://github.com/ansible-collections/community.docker/blob/f55a3f8154607a8913991f4d8f316b1f3ff02f3d/plugins/modules/docker_compose_v2.py#L539

They changed uppercase letters to lowercase letters and they removed some extra spaces in some messages. Also reporting of image pulling changed (but I do not support that for now, because I could tell the output wouldn't be stable). I've made parsing more lenient, but I still consider parsing "human targeted" output to be dirty.

plugins/modules/docker_compose_v2.py
felixfontein: That's fine, as long as it documents that it doesn't support them with `support: none` :) You really need to add this documentation fragment, otherwise the sanity tests will continue failing.

plugins/modules/docker_compose_v2.py
felixfontein: I would remove the default. I personally prefer explicit over implicit.

Once the default is gone, the first paragraph should be `When set to C(smart), contains will be recreated ...` or something like that anyway.

kwint:
What is the status about this? Anything we can help with?

Is there an issue on composes repo to ask for a stable output to stderr/stdout?

dansou901:
It's been a few months since the last status update. I'm currently restructuring my ansible collections / roles which also use docker compose, so I try to determine my best course of action. I can either use the shell module to run docker compose v2 or wait for the implementation of this module. Is this still being developed?

sebdanielsson:
> It's been a few months since the last status update. I'm currently restructuring my ansible collections / roles which also use docker compose, so I try to determine my best course of action. I can either use the shell module to run docker compose v2 or wait for the implementation of this module. Is this still being developed?

If you decide to go with the first option. Would you like to share your implementation?

dansou901:
My implementation for compose up -d is the following:
```yaml
- name: start container
  shell: docker compose --ansi never up -d --no-recreate
  args:
    executable: /bin/bash
    chdir: /path/to/docker-compose.yml 
  register: compose
  tags:
    - docker_compose
- name: check if container is running
  assert:
    that:
      - "'Running' in compose.stderr or 'Started' in compose.stderr" 
  tags:
    - docker_compose
```

dansou901:
If you want to loop that, it looks like
```yaml
- name: start containers
  shell: docker compose --ansi never up -d --no-recreate
  args:
    executable: /bin/bash
    chdir: "{{ item }}"
  loop:
    - /path1/to/docker-compose.yml
    - /path2/to/docker-compose.yml
    - ...
  register: compose
- name: check if containers are running
  assert:
    that:
      - "'Running' in item.stderr or 'Started' in item.stderr"
  loop: "{{ compose.results }}"
```

Callum027:
Here is what I've been using. It's fully idempotent (and therefore Molecule-friendly) without using `ansible.builtin.fail` or `ansible.builtin.assert`.

```yaml
- name: Create compose file
  ansible.builtin.template:
    src: docker-compose.yml.j2
    dest: /path/to/docker-compose.yml
    owner: root
    group: docker
    mode: "0644"

- name: Validate compose file
  ansible.builtin.command:
    cmd: docker compose --file docker-compose.yml config --quiet
    chdir: /path/to
  changed_when: false

- name: Pull container images
  ansible.builtin.command:
    cmd: docker compose --ansi never --progress plain --file docker-compose.yml pull
    chdir: /path/to
  register: temp_compose_pull
  changed_when: "'Downloading' in temp_compose_pull"

- name: Start containers
  ansible.builtin.command:
    cmd: docker compose --ansi never --progress plain --file docker-compose.yml up --detach
    chdir: /path/to
  register: temp_compose_up
  changed_when: "'Started' in temp_compose_up"
```

kaysond:
Bump. It's a real bummer that the compose plugin isn't supported yet since v1 has been deprecated for a while now.

Is there anything blocking this PR?

SirousSayahi:
Hi
Is this module still being developed? 
Compose v1 has been deprecated for a while now. It is strange that there is no update yet!

felixfontein:
@geerlingguy this PR is stale and merging my new PR (#739) will close it. If there are no big complaints I'm planning on merging my PR and getting a new release out by the end of this week.

lel-amri:
Hello @felixfontein,

Thanks for continuing this work. I have no complaints. I'd just like to reiterate that until Docker Compose's output gets stable, the method we are using for reading its state is very fragile (as a reminder, it changed at least two times while I was developing this PR). I kind of lost interest in this over time because of how hard it seems to me to stabilize Docker Compose internals in this regard (and also because existing issues didn't get a lot of attention).

I'm wondering what would be the impact on users if we advertise the docker_compose_v2 as stable as-is though. I think we should test this with different versions of docker-compose, explicitly list versions known to be compatible, and put a disclaimer about the uncertainty of compatibility with untested versions.

felixfontein:
> Thanks for continuing this work. I have no complaints. I'd just like to reiterate that until Docker Compose's output gets stable, the method we are using for reading its state is very fragile (as a reminder, it changed at least two times while I was developing this PR).

I fully agree. This is a serious problem with Docker Compose in general (it's also not surprising from how much feedback https://github.com/docker/compose/issues/10872 got since August).

> I kind of lost interest in this over time because of how hard it seems to me to stabilize Docker Compose internals in this regard (and also because existing issues didn't get a lot of attention).

(I guess you mean the above issue.)

> I'm wondering what would be the impact on users if we advertise the docker_compose_v2 as stable as-is though. I think we should test this with different versions of docker-compose, explicitly list versions known to be compatible, and put a disclaimer about the uncertainty of compatibility with untested versions.

Nothing in this collection is "stable" for all future versions of things they depend on (and that's pretty much impossible anyway). But I guess explicitly mentioning it for the new module is a good idea, since Docker Compose seems to be less stable than other things (looking at the breaking change https://github.com/docker/compose/pull/10918, for example, which got simply labelled as a "bugfix").

Right now CI tests with several versions (as you can see from the various `self.compose_version` comparisons in https://github.com/ansible-collections/community.docker/blob/d30085526a02b4ff931d7174eb7ebcea4aab6c00/plugins/modules/docker_compose_v2.py)

Since the actual docker compose plugin versions depend on various factors and change over times (especially for the rolling releases used in CI, like Arch Linux) I don't want to explicitly list the versions this is being tested with though. (If someone wants to automate retrieving the versions this is tested with and create tooling to create PRs which update the documentation I'll happily review and use that.)

Issue #611, opened 2023-04-28
Incompatibility between requests 2.29.0 and/or urllib3 2.0.0 and vendored Docker SDK for Python code


felixfontein:
##### SUMMARY
Currently there are a lot of CI failures. The problem seems to be urllib3 2.0.0, which has trouble with some requests versions: https://github.com/docker/docker-py/issues/3113

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
various

felixfontein:
```
02:33   File "/tmp/ansible_docker_container_copy_into_payload_xvaobvmd/ansible_docker_container_copy_into_payload.zip/ansible_collections/community/docker/plugins/module_utils/_api/api/client.py", line 236, in _put
02:33   File "/usr/local/lib/python3.10/dist-packages/requests/sessions.py", line 647, in put
02:33     return self.request("PUT", url, data=data, **kwargs)
02:33   File "/usr/local/lib/python3.10/dist-packages/requests/sessions.py", line 587, in request
02:33     resp = self.send(prep, **send_kwargs)
02:33   File "/usr/local/lib/python3.10/dist-packages/requests/sessions.py", line 701, in send
02:33     r = adapter.send(request, **kwargs)
02:33   File "/usr/local/lib/python3.10/dist-packages/requests/adapters.py", line 487, in send
02:33     resp = conn.urlopen(
02:33   File "/usr/local/lib/python3.10/dist-packages/urllib3/connectionpool.py", line 703, in urlopen
02:33     httplib_response = self._make_request(
02:33   File "/usr/local/lib/python3.10/dist-packages/urllib3/connectionpool.py", line 396, in _make_request
02:33     conn.request_chunked(method, url, **httplib_request_kw)
02:33 AttributeError: 'UnixHTTPConnection' object has no attribute 'request_chunked'
02:33 fatal: [testhost]: FAILED! => {
02:33     "changed": false,
[...]

PR #613, opened 2023-05-02
Make compatible with requests 2.29.0 and urllib3 2.0

felixfontein:
##### SUMMARY
Try to fix at least parts of #611.

Since requests 2.29.0, requests uses urllib3's native chunking ability (see https://github.com/psf/requests/pull/6226). By using urllib3's `HTTPConnection` class (which extends httplib.HTTPConnection`) we make sure that our HTTP transports support this as well.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
vendored Docker SDK for Python

### DIFF
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -101,6 +101,10 @@ jobs:
           - ''
         target:
           - ''
+        extra-constraints:
+          # Specifying this other than '' likely destroys change detection, but at least it will make
+          # CI pass when necessary...
+          - ''
         exclude:
           - ansible: ''
         include:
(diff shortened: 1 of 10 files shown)


felixfontein:
According to https://github.com/docker/docker-py/issues/3113#issuecomment-1531570788 this should also make the code compatible with urllib3 2.0.

felixfontein:
https://github.com/ansible-collections/community.docker/actions/runs/4874702225/jobs/8695973844?pr=614 shows that this is not done yet with urllib3 2.0. Unfortunately requests 2.30.0 has been yanked, so I cannot test this in CI anymore until they make a new release.

felixfontein:
Hmm, I didn't notice that the task that fails isn't our code, but pip. So this isn't a problem with this collection's code. I still think df8a4929ebfe7560ad99ad5ef015c0c2fda917ed is a good idea though.

Issue #605, opened 2023-04-10
Facts gathering hangs/errors when using TCP TLS socket


christophert:
##### SUMMARY

Hosts targeted via the `docker_containers` inventory module fail facts gathering if TCP TLS socket is in use.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

`community.docker.docker_containers.docker_api` connection
##### ANSIBLE VERSION

```paste below
ansible [core 2.14.4]
  config file = /home/user/Documents/Projects/ansible/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/Documents/Projects/ansible/.venv/lib64/python3.11/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/user/Documents/Projects/ansible/.venv/bin/ansible
  python version = 3.11.2 (main, Feb  8 2023, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)] (/home/user/Documents/Projects/ansible/.venv/bin/python)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
# /home/user/.ansible/collections/ansible_collections
Collection       Version
---------------- -------
community.docker 3.4.3  

# /home/user/Documents/Projects/ansible/.venv/lib64/python3.11/site-packages/ansible_collections
Collection       Version
---------------- -------
community.docker 3.4.3  

# /home/user/Documents/Projects/ansible/.venv/lib/python3.11/site-packages/ansible_collections
Collection       Version
---------------- -------
community.docker 3.4.3  
```

##### CONFIGURATION

```paste below
CONFIG_FILE() = /home/user/Documents/Projects/ansible/ansible.cfg
DEFAULT_ROLES_PATH(/home/user/Documents/Projects/ansible/ansible.cfg) = ['/home/user/Documents/Projects/ansible/roles']
```

##### OS / ENVIRONMENT

```
Linux vm-dev 6.2.9-200.fc37.aarch64 #1 SMP PREEMPT_DYNAMIC Thu Mar 30 22:54:14 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux
```
Docker host:
```
[...]

PR #621, opened 2023-05-12
Warn that SSLSocket cannot send close_notify TLS alerts

felixfontein:
##### SUMMARY
Ref: #605.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
changelog
docs

### DIFF
--- a/plugins/modules/docker_container_exec.py
+++ b/plugins/modules/docker_container_exec.py
@@ -95,7 +95,7 @@ options:
 
 notes:
   - Does not support C(check_mode).
-  - Does B(not work with TCP TLS sockets) when I(stdin) is used! This is caused by the inability to send C(close_notify) without closing the connection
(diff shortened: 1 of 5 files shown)


plugins/modules/docker_container_exec.py
```diff
@@ -95,6 +95,8 @@
 
 notes:
   - Does not support C(check_mode).
+  - Does B(not work with TCP TLS sockets) when I(stdin) is used! This is caused by the inability to send C(close_notify) without closing the connection
```
oraNod, line 98:
```suggestion
  - Does B(not work with TCP TLS sockets) when using I(stdin). This is caused by the inability to send C(close_notify) without closing the connection
```

oraNod (approved):

felixfontein:
@oraNod thanks a lot!

PR #645, opened 2023-06-11
Use semantic markup

felixfontein:
##### SUMMARY
Only to be merged once the last Ansible 7.x.y release is out.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
documentation


felixfontein:
As expected, the main problem here are the environment variables. (The missing envvar ref targets make the docs build fail.) For this collection we can circumvent this by explicitly adding targets, but that creates potential problems for the Ansible docsite, if any other collection defines targets for the *same* envvars.

felixfontein:
With antsibull-docs 2.1.0 CI now passes :)

felixfontein:
Restart CI.

PR #694, opened 2023-10-07
Update vendored Docker SDK for Python code

felixfontein:
##### SUMMARY
Vendor all relevant commits from https://github.com/docker/docker-py/compare/bc4c0d7cf4f6f794f9e92d93ddec02626eda739c...c38656dc7894363f32317affecc3e4279e1163f8.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
vendored Docker SDK for Python code

### DIFF
--- a/plugins/module_utils/_api/utils/socket.py
+++ b/plugins/module_utils/_api/utils/socket.py
@@ -15,7 +15,6 @@ import os
 import select
 import socket as pysocket
 import struct
-import sys
 
 from ansible.module_utils.six import PY3, binary_type
 
@@ -43,7 +42,7 @@ def read(socket, n=4096):
     recoverable_errors = (errno.EINTR, errno.EDEADLK, errno.EWOULDBLOCK)
 
     if PY3 and not isinstance(socket, NpipeSocket):
-        if sys.platform == 'win32':
+        if not hasattr(select, "poll"):
             # Limited to 1024
             select.select([socket], [], [])
         else:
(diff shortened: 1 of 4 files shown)


plugins/module_utils/_api/utils/socket.py
```diff
@@ -43,7 +42,7 @@ def read(socket, n=4096):
     recoverable_errors = (errno.EINTR, errno.EDEADLK, errno.EWOULDBLOCK)
 
     if PY3 and not isinstance(socket, NpipeSocket):
-        if sys.platform == 'win32':
+        if not hasattr(select, "poll"):
```
briantist, line 45: I don't really understand the specifics of what's going on here, but the conditional is quite different. Was the issue actually that `select.poll` was not defined on Windows platforms? If the attribute exists on Windows but did not work (or didn't work correctly for this situation) then this test would cause failures.

I guess, this code being vendored, it never will run on Windows (unless I ever get around to trying to get Ansible Python modules to execute on Windows but alas....).

Still, what was the motivation for this change?

felixfontein: This has been cherry-picked from https://github.com/docker/docker-py/commit/78439ebbe1aae77ff0fd9b666894d80807182e28 to support [eventlet](https://github.com/eventlet/eventlet). Our code neither supports Windows (outside WSL2) nor does it support eventlet, I'd still like the socket code to be as close as possible to Docker SDK for Python's code even for features we have no need for.

briantist (approved):

felixfontein:

PR #698, opened 2023-10-15
Add inventory filter capability

### LINKED ISSUE #610: Offer an include/exclude containers based on a filtering options in the Docker dynamic inventory
##### SUMMARY
It would be nice to offer the possibility to exclude certain containers from the dynamic inventory, via a white or black list parameter.

For example, in the `amazon.aws.aws_ec2` dynamic inventory, they are proposing two options: 
* [`exclude_filters`](https://docs.ansible.com/ansible/latest/collections/amazon/aws/aws_ec2_inventory.html#parameter-exclude_filters)
* [`include_filters`](https://docs.ansible.com/ansible/latest/collections/amazon/aws/aws_ec2_inventory.html#parameter-include_filters)

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`community.docker.docker_containers`; `plugins/inventory/docker_containers.py`

##### ADDITIONAL INFORMATION
It would really be nice to have the same kind of feature as the `aws_ec2` dynamic inventory on the Docker one, for example, to exclude; or only include containers with a certain label.

E.g.:
```yaml
plugin: community.docker.docker_containers
docker_host: unix://var/run/docker.sock

include_filters: 
  ## will only include containers having the custom label 'foo' in the inventory
  ## e.g.: In their Dockerfile,`LABEL my_customer_label=foo`
  - docker_config.Labels.my_custom_label:
    - foo 
```

felixfontein:
##### SUMMARY
Adds a `filter` option to all inventory plugins which allows to exclude/include hosts based on Jinja2 expressions.

This is a generic mechanism which can also be used in other inventory plugins (outside of this collection), similar to the `Constructable` mixin from ansible-core.

Fixes #610.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
inventory plugins


briantist (commented):
This is very cool! While it might be weird to take a dependency on `c.docker` for this alone, the plugin_util isn't named with an underscore or a notice about it being private. Would it be considered fair game to use it in that case? I don't have any plans to right now, just mentioning in case you didn't want to have to keep breaking changes to major versions or whatever.

felixfontein:
IMO using this plugin_utils / docs_fragment from community.docker in other collections would be fine. If you plan to do so (especially with public collections) please make sure you test against the current `main` branch of community.docker as well, to early on catch breaking changes should they ever arise (intentionally or not).

felixfontein:
The docs build error is expected as the workflow is run from the `main` branch.

felixfontein:
The dependency is now part of Ansible, so it's time to continue this :)

felixfontein:
recheck

felixfontein:
recheck

felixfontein:
Too bad if third-party-check fails...

PR #739, opened 2024-01-02
Add docker_compose_v2 module

### LINKED ISSUE #216: docker-compose 2.0.0 is no longer written in Python, thus breaking the docker_compose module

felixfontein:
##### SUMMARY
Implements a basic module wrapping the Docker CLI compose plugin.

I copied some of the output parsing code from #586. Unfortunately `docker compose`'s status reporting is pretty primitive, see also https://github.com/docker/compose/issues/10872.

Fixes #216.
Closes #586.

Reference for changes to compose over time: https://docs.docker.com/compose/release-notes/

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
docker_compose_v2

### DIFF
--- a/plugins/modules/docker_compose_v2.py
+++ b/plugins/modules/docker_compose_v2.py
@@ -1,0 +1,747 @@
+#!/usr/bin/python
+#
+# Copyright (c) 2023, Felix Fontein <felix@fontein.de>
+# Copyright (c) 2023, Léo El Amri (@lel-amri)
+# Copyright 2016 Red Hat | Ansible
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+
+module: docker_compose_v2
+
+short_description: Manage multi-container Docker applications with Docker Compose CLI plugin
+
+version_added: 3.6.0
+
+description:
+  - Uses Docker Compose to start or shutdown services.
+
+extends_documentation_fragment:
+  - community.docker.docker.cli_documentation
+  - community.docker.attributes
+  - community.docker.attributes.actiongroup_docker
+
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+
+options:
+  project_src:
+    description:
+      - Path to a directory containing a C(docker-compose.yml) or C(docker-compose.yaml) file.
+    type: path
+    required: true
+  project_name:
+    description:
+      - Provide a project name. If not provided, the project name is taken from the basename of O(project_src).
+    type: str
+  env_files:
+    description:
+      - By default environment files are loaded from a C(.env) file located directly under the O(project_src) directory.
+      - O(env_files) can be used to specify the path of one or multiple custom environment files instead.
+      - The path is relative to the O(project_src) directory.
+    type: list
+    elements: path
+  profiles:
+    description:
+      - List of profiles to enable when starting services.
+      - Equivalent to C(docker compose --profile).
+    type: list
+    elements: str
+  state:
+    description:
+      - Desired state of the project.
+      - V(present) is equivalent to running C(docker compose up).
+      - V(stopped) is equivalent to running C(docker compose stop).
+      - V(absent) is equivalent to running C(docker compose down).
+      - V(restarted) is equivalent to running C(docker compose restart).
+    type: str
+    default: present
+    choices:
+      - absent
+      - stopped
+      - restarted
+      - present
+  dependencies:
+    description:
+      - When O(state) is V(present) or V(restarted), specify whether or not to include linked services.
+    type: bool
+    default: true
+  recreate:
+    description:
+      - By default containers will be recreated when their configuration differs from the service definition.
+      - Setting to V(never) ignores configuration differences and leaves existing containers unchanged.
+      - Setting to V(always) forces recreation of all existing containers.
+    type: str
+    default: auto
+    choices:
+      - always
+      - never
+      - auto
+  remove_images:
+    description:
+      - Use with O(state=absent) to remove all images or only local images.
+    type: str
+    choices:
+      - 'all'
+      - 'local'
+  remove_volumes:
+    description:
+      - Use with O(state=absent) to remove data volumes.
+    type: bool
+    default: false
+  remove_orphans:
+    description:
+      - Remove containers for services not defined in the Compose file.
+    type: bool
+    default: false
+  timeout:
+    description:
+      - Timeout in seconds for container shutdown when attached or when containers are already running.
+    type: int
+
+requirements:
+  - "Docker CLI with Docker compose plugin 2.18.0 or later"
+
+author:
+  - Felix Fontein (@felixfontein)
+
+seealso:
+  - module: community.docker.docker_compose
+'''
+
+EXAMPLES = '''
+# Examples use the django example at https://docs.docker.com/compose/django. Follow it to create the
+# flask directory
+
+- name: Run using a project directory
+  hosts: localhost
+  gather_facts: false
+  tasks:
+    - name: Tear down existing services
+      community.docker.docker_compose_v2:
+        project_src: flask
+        state: absent
+
+    - name: Create and start services
+      community.docker.docker_compose_v2:
+        project_src: flask
+      register: output
+
+    - name: Show results
+      ansible.builtin.debug:
+        var: output
+
+    - name: Run `docker-compose up` again
+      community.docker.docker_compose_v2:
+        project_src: flask
+      register: output
+
+    - name: Show results
+      ansible.builtin.debug:
+        var: output
+
+    - ansible.builtin.assert:
+        that: not output.changed
+
+    - name: Stop all services
+      community.docker.docker_compose_v2:
+        project_src: flask
+        state: stopped
+      register: output
+
+    - name: Show results
+      ansible.builtin.debug:
+        var: output
+
+    - name: Verify that web and db services are not running
+      ansible.builtin.assert:
+        that:
+          - "not output.services.web.flask_web_1.state.running"
+          - "not output.services.db.flask_db_1.state.running"
+
+    - name: Restart services
+      community.docker.docker_compose_v2:
+        project_src: flask
+        state: restarted
+      register: output
+
+    - name: Show results
+      ansible.builtin.debug:
+        var: output
+
+    - name: Verify that web and db services are running
+      ansible.builtin.assert:
+        that:
+          - "output.services.web.flask_web_1.state.running"
+          - "output.services.db.flask_db_1.state.running"
+'''
+
+RETURN = '''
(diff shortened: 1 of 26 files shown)


felixfontein:
This is now ready for review!

plugins/modules/docker_compose_v2.py
```diff
@@ -0,0 +1,757 @@
...
+author:
+  - Felix Fontein (@felixfontein)
+
+notes:
+  - The Docker compose CLI plugin has no stable output format (see for example U(https://github.com/docker/compose/pull/10918)),
+    and for the main operations also no machine friendly output format. The module tries to accomodate this with various
+    version-dependent behavior adjustments and with testing older and newer versions of the Docker compose CLI plugin.
+
+    Currently the module is tested with multiple plugin versions between 2.18.1 and 2.23.3. The exact list of plugin versions
+    will change over time. New releases of the Docker compose CLI plugin can break this module at any time.
```
felixfontein, line 122: @lel-amri I've added the above note in response to https://github.com/ansible-collections/community.docker/pull/586#issuecomment-1874177305.

felixfontein: https://ansible-collections.github.io/community.docker/pr/739/docker_compose_v2_module.html#notes

lel-amri: Great !

lel-amri (approved):

tests/sanity/ignore-2.13.txt
```diff
@@ -1,2 +1,3 @@
 .azure-pipelines/scripts/publish-codecov.py replace-urlopen
+plugins/modules/docker_compose_v2.py validate-modules:return-syntax-error
```
briantist, line 2: what in this module triggers a return syntax error?

briantist (approved):
Looks great to me! Nice to finally have a v2 option. Shame there's no stable/machine-friendly output from the tool.

plugins/modules/docker_compose_v2.py
geerlingguy: Sadly, I think this is about as good as you can get there. At least we can take heart in the fact that if Docker were to change the command's UX too drastically, it would break people's brains in addition to the Ansible code, so hopefully they keep it pretty stable.

tests/sanity/ignore-2.13.txt
felixfontein: It's the `type: raw` IIRC. Some older versions might also barf on `choices` in return values.

felixfontein:
@lel-amri thanks for your work on this in #586!
@lel-amri @briantist @geerlingguy thanks for reviewing!

b-reich:
@felixfontein Many thanks for your PR. Is there still the possibility for a `docker compose pull` or `docker compose build --pull`?

felixfontein:
@b-reich right now no. I'm currently working on a `docker_compose_v2_pull` module which models `docker compose pull`. I originally wanted to have it done on last weekend, but parsing the `docker compose pull` output turned out to be more complex than I hoped for. I hope I can find some time this weekend to finish it.

About `docker compose build`: I think it's also a good idea to have that as a separate module, and I can try taking a look at it. It would be great if you could create a new issue for this (a proper feature request) and sketch out one or multiple scenarios in which you (or others) would use `docker compose build`.

b-reich:
> @b-reich right now no. I'm currently working on a `docker_compose_v2_pull` module which models `docker compose pull`. I originally wanted to have it done on last weekend, but parsing the `docker compose pull` output turned out to be more complex than I hoped for. I hope I can find some time this weekend to finish it.
> 
> 
> 
> About `docker compose build`: I think it's also a good idea to have that as a separate module, and I can try taking a look at it. It would be great if you could create a new issue for this (a proper feature request) and sketch out one or multiple scenarios in which you (or others) would use `docker compose build`.

Ahh I understand. If there are things I can support, please ping me. 
I will create an issue. 
Is there any advantages over creating an extra module over the v1 behaviour as parameter?

felixfontein:
> Ahh I understand. If there are things I can support, please ping me.

Thanks!

> Is there any advantages over creating an extra module over the v1 behaviour as parameter?

Short answer: look at docker_image and compare its usage to the new docker_image_* modules.

Long answer: it reduces the complexity, both of the code and of the user interface (the argument spec). This benefits both the user (less bugs, documentation easier to understand since the amount of options per module is small, you don't have to wonder whether option `force` now applies to pulling, starting, stopping, building, removing, or multiple of these, ...) and contributors (the code is simpler, there are no potentially surprising interactions between different 'modes' of the same module that you didn't think of, ...). Also it makes handling check mode easier, since it's easy to document which operations support check mode and which don't (push doesn't, pull only partially, build and tag support it fully - so what should the documentation for docker_image say? you don't want to read a wall of text and compare it to every of your docker_image tasks to figure out which one work fine in check mode and which don't).

There are some cases where a combined module makes things easier to express (like for docker_image: you can build, tag, and push the tagged image in one single task!). I *hope* that there are no cases where separate modules make something impossible or something common very inefficient. For `docker_compose_v2` and `docker_compose_v2_pull` (and of course for `docker_image_*`), I'm pretty sure that this is the case. For `docker_compose_v2_build` it's kind of a guess, that's why I ask for use-cases to hopefully notice if a separate module is a bad idea :)

felixfontein:
(Just look at https://docs.ansible.com/ansible/latest/collections/community/docker/docker_image_module.html#parameters and note that originally the options grouped in `build` were top-level options as well. Of course in the beginning there were a lot less of them, and it didn't look so bad, but then someone requests a new option `x` for building, then someone a new option `y`, and eventually you end up with 20 options just specificly for building images. Then you have five more for tagging, four for pulling, three for pushing, and you end up with a large soup of parameters. Obviously the numbers I mentioned here are a bit exaggerated, but they're not *that* far from the truth, especially if you look at `docker build --help` and start looking for things docker_image doesn't yet support.)

felixfontein:
PR for docker_compose_v2_pull module: #751

pgassmann:
the module currently does not have support for --diff.
But the return values contain a nice list of actions. 

@felixfontein Would it be possible to display those in diff mode? or are there strict conventions for diff output?

I now implemented  a workaround with a second task:


```yaml
- name: docker compose up ({{ webserver_domain }} in {{ webserver_path }})
  community.docker.docker_compose_v2:
    project_src: '{{ webserver_path }}'
    state: present
    remove_orphans: yes
    pull: "{{ webserver_docker_pull }}"
    build: "{{ webserver_docker_pull }}"
  register: webserver__register_compose

- name: print actions by compose
  vars:
    actions: "{{ webserver__register_compose.actions }}"
  debug:
    var: actions
  when: webserver__register_compose.changed
  changed_when: webserver__register_compose.changed
```

This generates the following output. also works in check mode:

```matlab
TASK [teamapps.general.webserver : docker compose up (www-dev.example.com) in /container/contao)] *******
changed: [website-dev1]

TASK [teamapps.general.webserver : print actions by compose] ***********
changed: [website-dev1] => 
  actions:
  - id: contao-sftp-1
    status: Starting
    what: container
```

felixfontein:
@pgassmann I guess the question is what you expect from diff mode :) I agree that the actions are better than nothing, but it's not really what I personally would expect (I would expect it to tell me which configuration changes trigger the actions - but that's not so easy to retrieve). Let's maybe create an issue and collect ideas.

goldyfruit:
@felixfontein what is the equivalent of the `files` argument with the new module? Previously I was doing like this:

```yaml
  community.docker.docker_compose:
    project_src: "{{ _ovos_installer_working_directory }}/{{ _ovos_installer_project_name }}/compose"
    project_name: "{{ _project_name }}"
    files: "{{ _composition_files }}"
    pull: true
    remove_orphans: "{{ ovos_installer_docker_compose_remove_orphans }}"
    remove_images: "{{ ovos_installer_docker_compose_remove_images }}"
    remove_volumes: "{{ ovos_installer_docker_compose_remove_volumes }}"
```

felixfontein:
@goldyfruit there is no equivalent at the moment. I've never used that feature nor saw it being used, so I didn't want to implement it in the first version to avoid screwing it up ;) Please create an issue (Feature Request) for it.

goldyfruit:
> @goldyfruit there is no equivalent at the moment. I've never used that feature nor saw it being used, so I didn't want to implement it in the first version to avoid screwing it up ;) Please create an issue (Feature Request) for it.

Feature request created. :+1:

felixfontein:
If someone is interested, #776 adds support for a `scale` option to docker_compose_v2.

pgassmann:
Ansible 9.2.0 was released yesterday, containing the new docker_compose_v2 modules: https://github.com/ansible-community/ansible-build-data/blob/main/9/CHANGELOG-v9.rst

BTW: Is there anywhere a better readable, rendered version of this Changelog for ansible-community?

felixfontein:
> BTW: Is there anywhere a better readable, rendered version of this Changelog for ansible-community?

@pgassmann not yet, I'm currently working of creating a MD version, I'll hopefully have time to get a first working version done on Friday. I'm currently running very low on spare time for any open source work, but from Friday on it will be better :)

(If you're interested: https://github.com/ansible-community/antsibull-changelog/pull/139)

In the meantime you can complain to GitHub which screwed up the RST preview of quite many changelogs and other RST documents (https://github.com/orgs/community/discussions/86715).

pgassmann:
@felixfontein don't apologize for not having much time for open source work. You are the most active and helpful contributor on the issues that affect me. It's telling, that you are the one who already implemented a a solution that fixes the issue in my question.
Thank you for all your efforts, I appreciate your support!

Shouldn't the changelog be part of the rendered website or a docs site like:
https://ansible-community-website.readthedocs.io/
or https://docs.ansible.com

felixfontein:
The Ansible changelog traditionally always was on git only and was never part of the docsite (docs.ansible.com); I don't know the reasons for that. One possible reason might be that the docsite build takes quite some time, so it's usually only updated after the actual release happens, which can be confusing since for some time (a few hours at most) the changelog wouldn't mention the latest version.

In any case, here's a markdown version of the Ansible 9 changelog: https://github.com/felixfontein/ansible-build-data/blob/markdown/9/CHANGELOG-v9.md (PR: https://github.com/ansible-community/ansible-build-data/pull/364). And here's the community.docker changelog as markdown: https://github.com/felixfontein/community.docker/blob/changelog/CHANGELOG.md (PR: https://github.com/ansible-collections/community.docker/pull/788).

briantist:
I think it would be worthwhile to publish the changelog if possible. Individual collections can already do this fairly easily if they want to with a symlink and a small change, see:
- https://docs.ansible.com/ansible/devel/collections/community/hashi_vault/docsite/CHANGELOG.html
- https://github.com/ansible-collections/community.hashi_vault/blob/main/docs/docsite/extra-docs.yml#L5

But I can't speak to what this process would be for the community package as a whole, nor how delayed its publishing might be.

The collection-level published changelog could be added with a small PR for anyone who's interested in doing so!

PR #758, opened 2024-01-15
feat: add docker-compose services support.

### LINKED ISSUE #757: [feature] docker_compose_v2 restart specific service
##### SUMMARY

It would be very useful to have same functionality as v1 version to be able to restart only specific service.

docker_compose (v1)
`Use with state present to restart all containers defined in the Compose file.
If services is defined, only the containers listed there will be restarted.`

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

docker_compose_v2

sbaerlocher:
##### SUMMARY
- This change allows users to manage specific services within a Docker Compose configuration. Instead of starting, stopping, or restarting all services in a Docker Compose file, users can now select a subset of services to apply the operation to. This enhances flexibility and control in container management.

Fixes #757.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- docker_compose_v2.py

##### ADDITIONAL INFORMATION
- With this extension to the Docker Compose module, users can now select specific services in their `docker-compose.yml` file and apply operations such as `up`, `stop`, `restart`, and `down` to them. This is particularly useful in complex setups where only certain services need to be restarted or stopped, without affecting the entire configuration.

```paste below
# Example: Starting specific services with the new 'services' parameter
- name: Start specific services
  docker_compose_v2:
    state: present
    services:
      - db
      - api
```
This example demonstrates how only the db and api services are started, instead of all services in the Compose file.


plugins/modules/docker_compose_v2.py
```diff
@@ -99,6 +99,11 @@
     description:
       - Timeout in seconds for container shutdown when attached or when containers are already running.
     type: int
+  services:
+    description:
+      - Specifies a subset of services to be targeted when the state is set to either V(present), V(stopped), or V
```
felixfontein, line 104:
```suggestion
      - Specifies a subset of services to be targeted.
```
There was something missing after `V`, and since this applies to all four possible values of `state`, it might be best to not mention that at all. WDYT?

plugins/modules/docker_compose_v2.py
```diff
@@ -399,6 +404,7 @@ def __init__(self, client):
         self.remove_volumes = parameters['remove_volumes']
         self.remove_orphans = parameters['remove_orphans']
         self.timeout = parameters['timeout']
+        self.services = parameters['services']
```
felixfontein, line 407: If you do:
```suggestion
        self.services = parameters['services'] or []
```
...

plugins/modules/docker_compose_v2.py
```diff
@@ -433,6 +439,9 @@ def get_up_cmd(self, dry_run, no_start=False):
             args.append('--no-start')
         if dry_run:
             args.append('--dry-run')
+        if self.services:
+            for service in self.services:
+                args.append(service)
```
felixfontein, line 444: ... you can simplify this (and the others below) to
```suggestion
        for service in self.services:
            args.append(service)
```

felixfontein (commented):
Thanks for your contribution!

felixfontein:
(BTW, you can ignore the failing `ansible/third-party-check`, it currently always fails.)

felixfontein (approved):

PR #776, opened 2024-01-23
docker_compose_v2: add scale option

felixfontein:
##### SUMMARY
Adds a `scale` option equivalent to `docker compose up --scale`.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
docker_compose_v2


felixfontein:
Right now it does not inform the user about changes when the count is reduced, due to Docker Compose itself not reporting that it shuts down containers: https://github.com/docker/compose/issues/11391

felixfontein:
> Right now it does not inform the user about changes when the count is reduced, due to Docker Compose itself not reporting that it shuts down containers: [docker/compose#11391](https://github.com/docker/compose/issues/11391)

Has been fixed in https://github.com/docker/compose/pull/11393 and should be available in the next Docker Compose release.

PR #799, opened 2024-02-15
Compose v2: don't emit waiting events for healthy containers

clinta:
The output of the compose command that is used for parsing events can contain multiple events for containers with health checks.  With one event status Waiting then a following event with the status Healthy. This causes the module to erroneously report changes.

This change ensures only the last event for each container is reported.

##### SUMMARY

`docker compose --ansi never --progress plain --project-directory <dir> --detach --no-color --quiet-pull --` will emit multiple events for containers with health checks. For example:

```
# docker compose --ansi never --progress plain --project-directory /srv/netbox-docker up --detach --no-color --quiet-pull --
 Container netbox-docker-redis-cache-1  Running
 Container netbox-docker-postgres-1  Running
 Container netbox-docker-redis-1  Running
 Container netbox-docker-netbox-1  Running
 Container netbox-docker-netbox-housekeeping-1  Running
 Container netbox-docker-netbox-worker-1  Running
 Container netbox-docker-netbox-1  Waiting
 Container netbox-docker-netbox-1  Waiting
 Container netbox-docker-netbox-1  Healthy
 Container netbox-docker-netbox-1  Healthy
```

This causes the module to report changes because it reports 2 containers with the Waiting status. The last four events are all for the same container, and only the last status should be reported.

To fix this, when a "Healthy" event is processed, any "Waiting" events for that same container are removed from the events list.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

community.docker.docker_compose_v2

##### ADDITIONAL INFORMATION

Use ansible to manage a compose definition with health checks.

```paste below

```

### DIFF
--- a/plugins/module_utils/compose_v2.py
+++ b/plugins/module_utils/compose_v2.py
@@ -303,6 +303,8 @@ def parse_events(stderr, dry_run=False, warn_function=None):
                 warn_missing_dry_run_prefix = True
         event = _extract_event(line)
         if event is not None:
+            if event.status == "Healthy":
+                events = [e for e in events if not (e.resource_id == event.resource_id and e.status == "Waiting")]
             events.append(event)
             if event.status in DOCKER_STATUS_ERROR:
                 error_event = event


plugins/module_utils/compose_v2.py
```diff
@@ -303,6 +303,8 @@ def parse_events(stderr, dry_run=False, warn_function=None):
                 warn_missing_dry_run_prefix = True
         event = _extract_event(line)
         if event is not None:
+            if event.status == "Healthy":
+                events = [e for e in events if not (e.resource_id == event.resource_id and e.status == "Waiting")]
```
felixfontein, line 307: Please do not remove events here. This function should return **all** events it finds and **not** contain any custom logic.

felixfontein (commented):
Thanks for your contribution!

Whether a change happened is detected by the `has_changes` function, and events are converted to actions by the `extract_actions` function. Both rely on `DOCKER_STATUS_WORKING` to determine whether an event is a change or not (`Waiting` is on that list).

Simply removing `Waiting` from `DOCKER_STATUS_WORKING` should fix this problem. It should be added to another list `DOCKER_STATUS_OTHER` though, which should be added to `DOCKER_STATUS`, to make sure that parsing `Waiting` isn't changed.

Finally please note that you need a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

clinta:
Removing `Waiting` from `DOCKER_STATUS_WORKING` will have the effect of not returning changes if the last status for the container is still `Waiting`. I'm not sure if this can happen, I'm guessing it can based on timeout values.

This seems like a potentially undesirable side-effect. Which is why I only make changes if `Waiting` is followed by a `Heathy` event for that container.

Is it okay to not return changes if the last event for a container is `Waiting`?

felixfontein:
If it's simply waiting, then no change happened? I don't see why it should report changed when encountering `Waiting` at all.

felixfontein:
I created #804 to simply stop treating `Waiting` as an action.

felixfontein:
Closing since #804 has been merged. Thanks for getting this started!

PR #803, opened 2024-02-22
docker_compose_v2: ignore some pull events

### LINKED ISSUE #802: docker_compose_v2: Pull always check behavior to ignore pulls
##### SUMMARY
I am using `pull: always` to ensure that docker images that have been rebuilt on the same tag are also always pulled.
But with the pull option set to always, the module will always report a change. I think we might be able to detect if it has actually pulled something or not via the output.

Ideally the step should only be marked as changed, when the containers were recreated. 
I saw that you, @felixfontein, already did something similar in #797, can we maybe introduce that to the docker_compose_v2 module too?

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
docker_compose_v2

Thanks! 
And huge thanks for migrating to compose v2! I could finally remove all the ugly workarounds for the python library installation <3 you're amazing.

felixfontein:
##### SUMMARY
Fixes #802.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
docker_compose_v2

### DIFF
--- a/changelogs/fragments/803-compose-v2-pull.yml
+++ b/changelogs/fragments/803-compose-v2-pull.yml
@@ -1,0 +1,3 @@
+bugfixes:
+  - "docker_compose_v2 - do not treat service-level pull events as changes to avoid incorrect ``changed=true`` return value of ``pull=always``
+    (https://github.com/ansible-collections/community.docker/issues/802, https://github.com/ansible-collections/community.docker/pull/803)."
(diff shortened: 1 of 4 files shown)


felixfontein:
TODO: potentially also make check mode behavior configurable.

felixfontein:
Hmm, that TODO is not implementable, as the behavior comes from Docker Compose:
```
" DRY-RUN MODE -  foobar Pulling ",
" DRY-RUN MODE -  foobar Pulled ",
" DRY-RUN MODE -  Container compose-foobar-1  Recreate",
" DRY-RUN MODE -  Container compose-foobar-1  Recreated",
" DRY-RUN MODE -  Container a756aaa76d8c_compose-foobar-1  Starting",
" DRY-RUN MODE -  Container a756aaa76d8c_compose-foobar-1  Started"
```

PR #815, opened 2024-03-12
Prevent RCE via inventory plugins

felixfontein:
##### SUMMARY
Fixing a potential RCE when using the inventory plugins.

Ref: https://www.die-welt.net/2024/03/remote-code-execution-in-ansible-dynamic-inventory-plugins/
Ref: https://forum.ansible.com/t/remote-code-execution-in-ansible-dynamic-inventory-plugins/4332/3

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
inventory plugins


felixfontein:
TIL: don't make `ansible_connection` unsafe. Some versions of ansible-core won't like it.

felixfontein:
(In this case it's OK anyway since that variable is set to a fixed string. So nothing nefarious can happen this way.)

markuman (approved):

felixfontein:
@markuman thanks again for reviewing this!

PR #843, opened 2024-05-01
docker_network: Add support for --config-from and --config-only

### LINKED ISSUE #395: docker_network: Support for --config-only and --from-config
### Summary

Recreating [this issue](https://github.com/ansible/ansible/issues/67819)

>Please support the setting "--config-only" and "--from-config" in docker_network
see docs (https://docs.docker.com/engine/reference/commandline/network_create/)

### Issue Type

Feature Idea

### Component Name

community.docker.docker_network

### Additional Information

```yaml (paste below)

```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

kmdm:
##### SUMMARY

Added support to docker_network.py for `--config-only` and `--config-from`

Fixes #395 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

docker_network

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/docker_network.py
+++ b/plugins/modules/docker_network.py
@@ -35,6 +35,18 @@ options:
     aliases:
       - network_name
 
+  config_from:
+    description:
+      - Specifies the config only network to use the config from.
+    type: str
+    version_added: 3.10.0
+
+  config_only:
+    description:
+      - Sets that this is a config only network.
+    type: bool
+    version_added: 3.10.0
+
   connected:
     description:
       - List of container names or container IDs to connect to a network.
@@ -283,6 +295,8 @@ class TaskParameters(DockerBaseClass):
 
         self.name = None
         self.connected = None
+        self.config_from = None
+        self.config_only = None
         self.driver = None
         self.driver_options = None
         self.ipam_driver = None
@@ -299,6 +313,11 @@ class TaskParameters(DockerBaseClass):
 
         for key, value in client.module.params.items():
             setattr(self, key, value)
+
+        # config_only sets driver to 'null' (and scope to 'local') so force that here. Otherwise we get
+        # diffs of 'null' --> 'bridge' given that the driver option defaults to 'bridge'.
+        if self.config_only:
+            self.driver = 'null'
 
 
 def container_names_in_network(network):
@@ -401,6 +420,14 @@ class DockerNetworkManager(object):
         :return: (bool, list)
         '''
         differences = DifferenceTracker()
+        if self.parameters.config_only is not None and self.parameters.config_only != net.get('ConfigOnly', False):
+            differences.add('config_only',
+                            parameter=self.parameters.config_only,
+                            active=net.get('ConfigOnly', False))
+        if self.parameters.config_from is not None and self.parameters.config_from != net.get('ConfigFrom', {}).get('Network', ''):
+            differences.add('config_from',
+                            parameter=self.parameters.config_from,
+                            active=net.get('ConfigFrom', {}).get('Network', ''))
         if self.parameters.driver and self.parameters.driver != net['Driver']:
             differences.add('driver',
                             parameter=self.parameters.driver,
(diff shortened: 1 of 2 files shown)


plugins/modules/docker_network.py
```diff
@@ -35,6 +35,17 @@
     aliases:
       - network_name
 
+  config_from:
+    description:
+      - Specifies the config only network to use the config from.
+    type: str
```
felixfontein, line 41:
```suggestion
    type: str
    version_added: 3.10.0
```

plugins/modules/docker_network.py
```diff
@@ -35,6 +35,17 @@
...
+
+  config_only:
+    description:
+      - Sets that this is a config only network.
+    type: bool
+    default: false
```
felixfontein, line 47: There should be no `default` to avoid unnecessarily changing network configs that were not created/adjusted with this module.
```suggestion
    version_added: 3.10.0
```

plugins/modules/docker_network.py
```diff
@@ -503,6 +528,10 @@ def create_network(self):
                 'CheckDuplicate': None,
             }
 
+            if self.parameters.config_only:
+                data['ConfigOnly'] = True
+            if self.parameters.config_from:
+                data['ConfigFrom'] = {'Network': self.parameters.config_from}
```
felixfontein, line 534: Interestingly, neither of these two parameters are documented: https://docs.docker.com/engine/api/v1.44/#tag/Network/operation/NetworkCreate
The Docker SDK for Python does not seem to provide these options at all, but the CLI passes them on using exactly these structures. I guess they simply forgot to document them in the OpenAPI spec :shrug:

plugins/modules/docker_network.py
```diff
@@ -300,6 +313,10 @@ def __init__(self, client):
         for key, value in client.module.params.items():
             setattr(self, key, value)
 
+        # config_only sets driver to 'null' so force that here otherwise we get
+        # diffs of 'null' --> 'bridge' given driver defaults to 'bridge'.
```
felixfontein, line 317:
```suggestion
        # config_only sets driver to 'null' (and scope to 'local') so force that here. Otherwise we get
        # diffs of 'null' --> 'bridge' given that the driver option defaults to 'bridge'.
```

felixfontein (commented):
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

plugins/modules/docker_network.py
```diff
@@ -630,6 +659,8 @@ def absent(self):
 def main():
     argument_spec = dict(
         name=dict(type='str', required=True, aliases=['network_name']),
+        config_from=dict(type='str', default=''),
+        config_only=dict(type='bool', default=False),
```
felixfontein, line 663:
```suggestion
        config_from=dict(type='str'),
        config_only=dict(type='bool'),
```

plugins/modules/docker_network.py
```diff
@@ -503,6 +528,10 @@ def create_network(self):
                 'CheckDuplicate': None,
             }
 
+            if self.parameters.config_only:
+                data['ConfigOnly'] = True
```
felixfontein, line 532:
```suggestion
            if self.parameters.config_only is not None:
                data['ConfigOnly'] = self.parameters.config_only
```

plugins/modules/docker_network.py
felixfontein: https://github.com/moby/moby/issues/47790

kmdm: I didn't even check - I took a fairly naive approach of mirroring the output of `docker network inspect` when I saw create operation appeared to be providing the fields in that format. 

I'll apply your changes, the changelog fragment and fixup/squash when I get a little time! 

Thanks for the review.

kmdm:
Added changelog fragment and accepted/squashed all suggested changes.

felixfontein (approved):
If nobody objects, I'll merge this on the weekend.

felixfontein:
@kmdm thanks for your contribution!

PR #847, opened 2024-05-08
Allow healthcheck override without test option

x4rd0o1Vtx:
##### SUMMARY

- Use case

Change healthcheck container options while keeping original image test.

ex:

```
healthcheck:
  interval: 1m
  timeout: 10s
  start_period: 30s
  retries: 3      
```

- Actual behavior

```
"Healthcheck" {
  "Test": ["NONE"]
}
```

- New behavior

```
"Healthcheck" {
  "Test": <internal docker image test>,
  "Interval": 60000000000,
  "Timeout": 10000000000,
  "StartPeriod": 30000000000,
  "Retries": 3
}
```

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

docker_api

### DIFF
--- a/plugins/module_utils/module_container/base.py
+++ b/plugins/module_utils/module_container/base.py
@@ -935,6 +935,7 @@ OPTION_HEALTHCHECK = (
     OptionGroup(preprocess=_preprocess_healthcheck)
     .add_option('healthcheck', type='dict', ansible_suboptions=dict(
         test=dict(type='raw'),
+        test_cli_compatible=dict(type='bool', default=False),
         interval=dict(type='str'),
         timeout=dict(type='str'),
         start_period=dict(type='str'),
--- a/plugins/module_utils/module_container/docker_api.py
+++ b/plugins/module_utils/module_container/docker_api.py
@@ -746,7 +746,7 @@ def _preprocess_etc_hosts(module, client, api_version, value):
 def _preprocess_healthcheck(module, client, api_version, value):
     if value is None:
         return value
-    if not value or not value.get('test'):
+    if not value or not (value.get('test') or (value.get('test_cli_compatible') and value.get('test') is None)):
         value = {'test': ['NONE']}
     elif 'test' in value:
         value['test'] = normalize_healthcheck_test(value['test'])
--- a/plugins/module_utils/util.py
+++ b/plugins/module_utils/util.py
@@ -353,7 +353,7 @@ def normalize_healthcheck(healthcheck, normalize_test=False):
     result = dict()
 
     # All supported healthcheck parameters
-    options = ('test', 'interval', 'timeout', 'start_period', 'start_interval', 'retries')
+    options = ('test', 'test_cli_compatible', 'interval', 'timeout', 'start_period', 'start_interval', 'retries')
 
     duration_options = ('interval', 'timeout', 'start_period', 'start_interval')
 
@@ -366,7 +366,7 @@ def normalize_healthcheck(healthcheck, normalize_test=False):
                 continue
             if key in duration_options:
                 value = convert_duration_to_nanosecond(value)
-            if not value:
+            if not value and not (healthcheck['test_cli_compatible'] and key == 'test'):
                 continue
             if key == 'retries':
                 try:
@@ -376,7 +376,7 @@ def normalize_healthcheck(healthcheck, normalize_test=False):
                         'Cannot parse number of retries for healthcheck. '
                         'Expected an integer, got "{0}".'.format(value)
                     )
-            if key == 'test' and normalize_test:
+            if key == 'test' and value and normalize_test:
                 value = normalize_healthcheck_test(value)
             result[key] = value
 
--- a/plugins/module_utils/util.py
+++ b/plugins/module_utils/util.py
@@ -366,7 +366,7 @@ def normalize_healthcheck(healthcheck, normalize_test=False):
                 continue
             if key in duration_options:
                 value = convert_duration_to_nanosecond(value)
-            if not value and not (healthcheck['test_cli_compatible'] and key == 'test'):
+            if not value and not (healthcheck.get('test_cli_compatible') and key == 'test'):
                 continue
             if key == 'retries':
                 try:
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -388,6 +388,7 @@ options:
             This is the classical behavior of the module and currently the default behavior.
         default: false
         type: bool
+        version_added: 3.10.0
       interval:
         description:
           - Time between running the check.
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -379,6 +379,15 @@ options:
           - Command to run to check health.
           - Must be either a string or a list. If it is a list, the first item must be one of V(NONE), V(CMD) or V(CMD-SHELL).
         type: raw
+      test_cli_compatible:
+        description:
+          - If set to V(true), omitting O(healthcheck.test) while providing O(healthcheck) does not disable healthchecks,
+            but simply overwrites the image's values by the ones specified in O(healthcheck). This is
+            the behavior used by the Docker CLI.
+          - If set to V(false), omitting O(healthcheck.test) will disable the container's health check.
+            This is the classical behavior of the module and currently the default behavior.
+        default: false
+        type: bool
       interval:
         description:
           - Time between running the check.
--- a/changelogs/fragments/847-docker_container-heackcheck-test_cli_compatible.yml
+++ b/changelogs/fragments/847-docker_container-heackcheck-test_cli_compatible.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - docker_container - adds ``healthcheck.test_cli_compatible`` to allow omit test option on containers without remove existing image test (https://github.com/ansible-collections/community.docker/pull/847).


felixfontein (commented):
Thanks for your contribution! Unfortunately this is a breaking change.

Maybe this could be done instead by adding a new option `keep_image_test` (type boolean, default `false`), which when set to `true` ignores the `test` value?

x4rd0o1Vtx:
Thanks to you for maintaining this module! Indeed, I realised that few moment later after creating this PR...

I like this way to keep closest to docker cli behavior and this allow to do both keeping original test image by default or explicitly set ["NONE"] to unset it. Moreover, I saw there is only one rarely case to get the breaking change.

- Still keep the same behavior

```
healthcheck: {}
```

```
healthcheck:
  test: "" (or false, {}, [], omit)
```

=> "Healthcheck" {"Test": ["NONE"]}

- But not this one

```
healthcheck:
  test: "" (or false, {}, [], omit)
  interval: 1m
```

=> "Healthcheck" {"Test": ["NONE"]} (actual)
=> "Healthcheck" {"Test": None} (new)

This makes no sense to me because why adding test options like `interval` if you just want to unset the image test. But maybe this can result from conditional on test option follow by other options. So in this particular case, user should explicitly set ["NONE"]  to test option to keep the actual behavior.

To handle this, maybe first reduce the breaking area to the omit case only, like I try in this commit by forwarding the real test value. (not 100% sure of potential side effect)

And if it's not enougth, maybe a `mode` or `cli_compatible` option could be a good way to keep the actual behavior and allow a breaking change on it on the future if you wish.

What do you think ?

x4rd0o1Vtx:
This seems to have unexpected side effect based on some failing tests, I'll wait your feedback on my previous comment before going deep on it.

felixfontein:
> This makes no sense to me because why adding test options like `interval` if you just want to unset the image test. But maybe this can result from conditional on test option follow by other options. So in this particular case, user should explicitly set ["NONE"] to test option to keep the actual behavior.
> 
> To handle this, maybe first reduce the breaking area to the omit case only, like I try in this commit by forwarding the real test value. (not 100% sure of potential side effect)

I agree that it doesn't make sense, but it's still a breaking change of a behavior that has been this way for quite a few years. Someone might rely on this, and suddenly their deployments change (potentially in a fatal way).

Such changes need to come with a longer deprecation period where warnings are shown to the users. Since nobody wants to wait years to be able to use the new behavior, the best course of action is usually to add a new option to control this behavior which defaults to the current behavior, then eventually deprecate the default behavior (showing deprecation warnings in situation this results in a change), and eventually change the default.

> And if it's not enougth, maybe a `mode` or `cli_compatible` option could be a good way to keep the actual behavior and allow a breaking change on it on the future if you wish.

Such an option would be best. A general `cli_compatible` option is not a good idea for this, since that behavior will also change over time when other incompatibilities are found. (And `mode` is way too generic, and usually means mode of a file.)

Having something inside the `healthcheck` dictionary is probably best, maybe an option called `test_cli_compatible`? That should also avoid potential collisions with things Docker adds, and it makes clear it is both about `test` and compatbility with the CLI.

x4rd0o1Vtx:
Thanks for taking the time to explain why this way is still not possible and I understand better when you immediately suggested adding a new option. I thought there were special cases especially when this one seems to me very closer to a bugfix, even the documentation example may suggest that you absolutely should set ["NONE"] to the test option to remove the container healthcheck and that omitting this value should not have any effect.

```
- name: Remove healthcheck from container
    name: nginx-proxy
    image: nginx:1.13
    state: started
    healthcheck:
      # The "NONE" check needs to be specified
      test: ["NONE"]
```

In reality, all of these unset the healtcheck:

```
healthcheck: {}
```

```
healthcheck:
  test: "" (or false, {}, [], omit)
```

However I'm not doing this for myself but for the community and you know better than me what's best for the project so I'm totally fine to add a `test_cli_compatible` option.

Last question before making the changes, should I keep only the initial commit behavior or it's better to reduce to the omit case only like my second commit did ?

- First commit behavior

```
healthcheck:
  test_cli_compatible: true
  test: "" (or false, {}, [], omit)
  interval: 1m
```
=> "Healthcheck" {"Test": None}

- Second commit behavior

```
healthcheck:
  test_cli_compatible: true
  test: "" (or false, {}, [])
  interval: 1m
```
=> "Healthcheck" {"Test": ["NONE"]} (still keep the actual behavior)

```
healthcheck:
  test_cli_compatible: true
  test: "{{ omit }}"
  interval: 1m
```
=> "Healthcheck" {"Test": None}

felixfontein:
I would use the second commit's behavior. If the user explicitly provides a value for `test`, it should override the image's value. (Unless `test` is not provided, or equivalently, if `{{ omit }}` is provided.)

x4rd0o1Vtx:
Seems good for me but not sure about some failing tests, could you review this again please. Thanks in advance for your time.

plugins/module_utils/module_container/base.py
```diff
@@ -935,6 +935,7 @@ def _compare_platform(option, param_value, container_value):
     OptionGroup(preprocess=_preprocess_healthcheck)
     .add_option('healthcheck', type='dict', ansible_suboptions=dict(
         test=dict(type='raw'),
+        test_cli_compatible=dict(type='bool'),
```
felixfontein, line 938: You don't have a default here:
```suggestion
        test_cli_compatible=dict(type='bool', default=False),
```

plugins/module_utils/module_container/docker_api.py
```diff
@@ -746,10 +746,11 @@ def _preprocess_etc_hosts(module, client, api_version, value):
 def _preprocess_healthcheck(module, client, api_version, value):
     if value is None:
         return value
-    if not value or not value.get('test'):
+    if not value or not (value.get('test') or (value.get('test_cli_compatible') and value.get('test') is None)):
         value = {'test': ['NONE']}
-    elif 'test' in value:
+    elif value.get('test'):
```
felixfontein, line 751: For `test_cli_compatible=false` this should stay `'test' in value`. So maybe:
```suggestion
    elif (value.get('test') if value.get('test_cli_compatible') else 'test' in value):
```
What's the reason to change it for `test_cli_compatible=true`?

plugins/module_utils/util.py
```diff
@@ -366,7 +366,7 @@ def normalize_healthcheck(healthcheck, normalize_test=False):
                 continue
             if key in duration_options:
                 value = convert_duration_to_nanosecond(value)
-            if not value:
+            if not value and key != 'test':
                 continue
```
felixfontein, line 370: This is problematic, as it modifies behavior even if `test_cli_compatible` is not set. Please change this so it does **not** modify any behavior if `test_cli_compatible` is not explicitly `True`.

Same below.

plugins/modules/docker_container.py
```diff
@@ -379,6 +379,11 @@
           - Command to run to check health.
           - Must be either a string or a list. If it is a list, the first item must be one of V(NONE), V(CMD) or V(CMD-SHELL).
         type: raw
+      test_cli_compatible:
+        description:
+          - Don't remove test container if test option is omit.
```
felixfontein, line 384:
```suggestion
          - If set to V(true), omitting O(healthcheck.test) while providing O(healthcheck) does not disable healthchecks,
            but simply overwrites the image's values by the ones specified in O(healthcheck). This is
            the behavior used by the Docker CLI.
          - If set to V(false), omitting O(healthcheck.test) will disable the container's health check.
            This is the classical behavior of the module and currently the default behavior.
```

plugins/module_utils/module_container/docker_api.py
x4rd0o1Vtx: No reason I believed there was the test key with None value but it's not.

plugins/module_utils/util.py
x4rd0o1Vtx: Absolutely, I've changed that.

plugins/module_utils/util.py
```diff
@@ -366,7 +366,7 @@ def normalize_healthcheck(healthcheck, normalize_test=False):
...
             if key in duration_options:
                 value = convert_duration_to_nanosecond(value)
-            if not value:
+            if not value and not (healthcheck['test_cli_compatible'] and key == 'test'):
```
felixfontein, line 369:
```suggestion
            if not value and not (healthcheck.get('test_cli_compatible') and key == 'test'):
```
`test_cli_compatible` is generally not there; only if this is called from the docker_container module.

plugins/module_utils/util.py
```diff
@@ -376,7 +376,7 @@ def normalize_healthcheck(healthcheck, normalize_test=False):
                         'Cannot parse number of retries for healthcheck. '
                         'Expected an integer, got "{0}".'.format(value)
                     )
-            if key == 'test' and normalize_test:
+            if key == 'test' and value and normalize_test:
```
felixfontein, line 379: This one also needs to be adjusted.

plugins/modules/docker_container.py
```diff
@@ -379,6 +379,15 @@
...
         type: raw
+      test_cli_compatible:
+        description:
+          - If set to V(true), omitting O(healthcheck.test) while providing O(healthcheck) does not disable healthchecks,
+            but simply overwrites the image's values by the ones specified in O(healthcheck). This is
+            the behavior used by the Docker CLI.
+          - If set to V(false), omitting O(healthcheck.test) will disable the container's health check.
+            This is the classical behavior of the module and currently the default behavior.
+        default: false
+        type: bool
```
felixfontein, line 390:
```suggestion
        type: bool
        version_added: 3.10.0
```

plugins/module_utils/util.py
x4rd0o1Vtx: Not really familiar with python language I didn't know that will raise an error.

plugins/module_utils/util.py
x4rd0o1Vtx: I don't have this one, for me that code still never reach due to previous continue if value missing (keep the current behavior) and the only case I see is when test_cli_compatible is true and forward an empty test value.
Did I miss something ?

felixfontein: No, everything's fine, I was too overcautious and didn't look closely enough :) The old code had `if not value: continue` further above, so in the old code path this `if` is never reached if `value` isn't truish.

felixfontein (approved):
If you want you can also add a similar change to the `docker_swarm_service` module. But maybe let's get this merged first so it's definitely part of the next release (which happens this weekend) :)

felixfontein:
@x4rd0o1Vtx thank you very much for your contribution!

plugins/module_utils/util.py
x4rd0o1Vtx: No problem always better to be overcautious than not enough ;)

x4rd0o1Vtx:
My pleasure and thanks for your time again !

Issue #860, opened 2024-05-20
requests==2.32.0 breakage: Not supported URL scheme http+docker


ssbarnea:
##### SUMMARY
Apparently `requests==2.32.0` which was released few hours ago is not compatible with docker.

https://github.com/psf/requests/releases/tag/v2.32.0

```
TASK [Create a container] ******************************************************
failed: [localhost] (item={'image': 'ubuntu:18.04', 'name': 'molecule-ubuntu'}) => {"ansible_loop_var": "item", "changed": false, "item": {"image": "ubuntu:18.04", "name": "molecule-ubuntu"}, "msg": "Error connecting: Error while fetching server API version: Not supported URL scheme http+docker"}
```

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```paste below
ansible [core 2.16.6]
  config file = None
  configured module search path = ['/Users/ssbarnea/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/ssbarnea/.asdf/installs/python/3.12.3/lib/python3.12/site-packages/ansible
  ansible collection location = /Users/ssbarnea/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/ssbarnea/.asdf/installs/python/3.12.3/bin/ansible
  python version = 3.12.3 (main, May 10 2024, 13:04:47) [Clang 15.0.0 (clang-1500.3.9.4)] (/Users/ssbarnea/.asdf/installs/python/3.12.3/bin/python3.12)
  jinja version = 3.1.4
  libyaml = True
```

##### COLLECTION VERSION

```paste below
$ ansible-galaxy collection list community.docker

# /Users/ssbarnea/.ansible/collections/ansible_collections
Collection       Version
---------------- -------
community.docker 3.10.0 
```

##### CONFIGURATION

```paste below
CONFIG_FILE() = None
DEFAULT_DEBUG(env: ANSIBLE_DEBUG) = False
DEVEL_WARNING(env: ANSIBLE_DEVEL_WARNING) = False
EDITOR(env: EDITOR) = code -r -a
PAGER(env: PAGER) = less
```

##### OS / ENVIRONMENT

macos aarch64

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```paste below

```

felixfontein:
There is no commit that could change that. I'm pretty sure this is not related to this collection, but to some third-party code.

[...]

PR #861, opened 2024-05-20
[DNM] Trigger all tests

felixfontein:
##### SUMMARY
Ref: #860

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
???


ssbarnea (approved):

felixfontein:
The failures in CI groups 1, 2, 3 (these are for Docker Swarm, using Docker SDK for Python) and 6 (docker-compose v1, also using Docker SDK for Python) are to be expected.

The late failures in CI group 4 are unfortunately also expected, since the docker_network tests (which generally don't use Docker SDK for Python) use Docker Swarm for one specific test (and for that it needs Docker SDK for Python).

(If one of these groups doesn't fail, it's likely because they have an older requests version :) )

felixfontein:
@ssbarnea thanks for reporting and reviewing this!

PR #899, opened 2024-06-28
Create helper OCI images for use in CI

felixfontein:
##### SUMMARY
We can then use them in CI as light-weight alternatives for some of the images we are using.

This is in particular important as the layers of these images are definitely not used by any other image used in CI (for example Alpine layers used by ansible-test images), so we can get rid of some of the annoying 'random CI failures' due to layers being still in use by running containers.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
The failing CI jobs show why these images will be useful...

PR #900, opened 2024-06-28
CI: use new container images for Compose v2 pull tests

felixfontein:
##### SUMMARY
This hopefully makes these tests a lot more stable in CI.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
docker_compose_v2 tests
docker_compose_v2_pull tests


felixfontein:
It turns out that docker compose has some special code that when pulling, it **ignores** that images with tag `latest` are already there locally if pull policy is `missing`: https://github.com/docker/compose/pull/9781 (changelog: https://docs.docker.com/compose/release-notes/#2101)

That's pretty unexpected... (And it would be really helpful if such a change would contain some justification.)

PR #915, opened 2024-07-06
fix mis-named keys and invalid values in copy into example

ethanjjjjjjj:
##### SUMMARY

Updated keys and values used in the Docker copy into container module examples, as they are invalid in the latest version of the module and no not match the spec above.

- `owner_id` and `group_id` should be used instead of `owner` and `group` respectively
- mode is expecting an integer, yet `0o755` is interpreted as a string

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

community.docker.docker_container_copy_into module

##### ADDITIONAL INFORMATION

```paste below

```


felixfontein (approved):
Thanks a lot for fixing the docs!

PR #921, opened 2024-07-07
docker_container: allow to wait for a container to become healthy

### LINKED ISSUE #890: state: healthy for docker_container
##### SUMMARY
Currently it's only possible to use state `absent`, `present`, `stopped` and `started` for the docker_container module. I would like to propose adding `state: healthy` to have Ansible wait until the integrated HEALTHCHECK returns healthy (when defined in the Dockerfile).

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
community.docker.docker_container

##### ADDITIONAL INFORMATION
By implementing this feature, a playbook can halt execution until the docker_container healtcheck returns healthy. This prevents additional tasks to poll for the health manually.

```yaml
- name: Create database container
  docker_container:
    name: "database"
    image: "mysql:latest"
    state: "healthy" # This will trigger Ansible to wait for the HEALTHCHECK statement to return healthy, and fail if unhealthy.
    restart: true
    detach: true
    pull: "always"
    restart_policy: always
    published_ports:
      - "{{ db_fixture_port }}:3306"
```

felixfontein:
##### SUMMARY
Fixes #890.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
docker_container

### DIFF
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -944,10 +944,10 @@ options:
     type: str
   healthy_wait_timeout:
     description:
-      - When waiting for the container to become healthy if O(state=healthy), this option controls for how long
-        the module waits until the container's state becomes healthy.
+      - When waiting for the container to become healthy if O(state=healthy), this option controls how long
+        the module waits until the container state becomes healthy.
       - The timeout is specified in seconds. The default, V(300), is 5 minutes.
-      - Set this to 0 or a negative value to wait however long it takes.
+      - Set this to 0 or a negative value to wait indefinitely.
         Note that depending on the container this can result in the module not terminating.
     default: 300
     type: float
--- a/plugins/modules/docker_container.py
+++ b/plugins/modules/docker_container.py
@@ -381,6 +381,7 @@ options:
       - "O(healthcheck.interval), O(healthcheck.timeout), O(healthcheck.start_period), and O(healthcheck.start_interval) are specified as durations.
         They accept duration as a string in a format that look like: V(5h34m56s), V(1m30s), and so on.
         The supported units are V(us), V(ms), V(s), V(m) and V(h)."
+      - See also O(state=healthy).
     type: dict
     suboptions:
       test:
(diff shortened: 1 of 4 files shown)


plugins/modules/docker_container.py
```diff
@@ -932,12 +935,23 @@
...
       - stopped
       - started
   stop_signal:
     description:
       - Override default signal used to stop the container.
     type: str
+  healthy_wait_timeout:
+    description:
+      - When waiting for the container to become healthy if O(state=healthy), this option controls for how long
+        the module waits until the container's state becomes healthy.
```
oraNod, line 948:
```suggestion
      - When waiting for the container to become healthy if O(state=healthy), this option controls how long
        the module waits until the container state becomes healthy.
```

plugins/modules/docker_container.py
```diff
@@ -932,12 +935,23 @@
...
+      - The timeout is specified in seconds. The default, V(300), is 5 minutes.
+      - Set this to 0 or a negative value to wait however long it takes.
```
oraNod, line 950:
```suggestion
      - Set this to 0 or a negative value to wait indefinitely.
```

plugins/modules/docker_container.py
```diff
@@ -919,6 +920,8 @@
         with the requested config.'
       - 'V(started) - Asserts that the container is first V(present), and then if the container is not running moves it to a running
         state. Use O(restart) to force a matching container to be stopped and restarted.'
+      - V(healthy) - Asserts that the container is V(present) and V(started), and is actually healthy as well.
```
oraNod, line 923:
```suggestion
      - V(healthy) - Asserts that the container is V(present) and V(started), and is running without errors.
```

oraNod: Not sure if that's the most accurate description of the healthy state but it avoids using the word "healthy" to describe the state "healthy".

oraNod (approved):
Some minor wording nits but the rest LGTM. Thanks @felixfontein

plugins/modules/docker_container.py
felixfontein: Hmm, I'm a bit hesitant about this, since `healthy` is the terminology for Docker containers (https://docs.docker.com/reference/dockerfile/#healthcheck).

How about the following:
```suggestion
      - V(healthy) - Asserts that the container is V(present) and V(started), and is actually healthy as well.
        This means that the conditions defined in O(healthcheck) respectively in the image's C(HEALTHCHECK)
        (L(Docker reference for HEALTHCHECK, https://docs.docker.com/reference/dockerfile/#healthcheck))
        are satisfied.
```

oraNod: @felixfontein Ah, I see. I think that suggestion is a good improvement for sure.

felixfontein:
@oraNod thanks a lot for reviewing this!

PR #922, opened 2024-07-07
docker_container: prevent infinite loop if removal_wait_timeout is specified

felixfontein:
##### SUMMARY
Found this while working on #921. The new use of `wait_for_state()` in that PR resulted in infinite loops in the tests without this fix when `max_wait=1` and the state would never be reached.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
docker_container


felixfontein:
CI failures are unrelated.

PR #933, opened 2024-07-19
docker_container: pass networks to Daemon on container creation

felixfontein:
##### SUMMARY
This works around https://github.com/moby/moby/issues/48192 at least during container creation. When later updating the networks it doesn't help, unfortunately.

Ref: https://forum.ansible.com/t/setting-a-mac-address-in-a-docker-container/7315/6

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
docker_container

### DIFF
--- a/changelogs/fragments/933-docker_container-networks.yml
+++ b/changelogs/fragments/933-docker_container-networks.yml
@@ -1,7 +1,11 @@
 minor_changes:
-  - "docker_container - when creating a network, directly pass the networks to connect to to the Docker Daemon.
-     This makes creation more efficient and works around a bug in Docker Daemon that does not use the specified
-     MAC address in at least some cases, though only for creation (https://github.com/ansible-collections/community.docker/pull/933)."
+  - "docker_container - when creating a container, directly pass all networks to connect to to the Docker Daemon
+     for API version 1.44 and newer. This makes creation more efficient and works around a bug in Docker Daemon that
+     does not use the specified MAC address in at least some cases, though only for creation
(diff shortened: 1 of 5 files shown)


felixfontein:
This seems to fail when passing more than one network. When looking at how Docker SDK for Python does this, it explicitly says
```
        You can specify networks to connect the container to by using the
        ``networking_config`` parameter. At the time of creation, you can
        only connect a container to a single networking, but you
        can create more connections by using
        :py:meth:`~connect_container_to_network`.
```
which kind of reminds me on that I was looking into this before. In fact, check out https://github.com/ansible/ansible/commit/4fac91bed51ec06a16c022b758af4739b09c638d#diff-5e7c531fbb5041f2e04c666990379baac0186369a2bf47fa6df4310c92365b9aR1287. This part of the code apparently got lost when switching to talking to the API directly (77e63e2cca44534a7ad0777f0ef97d4280bb62ff).

Even if we change back to that, it means that only one of the multiple networks will get specified on creation, and the others are attached later on. Which apparently does not result in the MAC address being used. But then, Compose can still do it for more than one network. So there's more work to do to figure out what's happening there...

felixfontein:
Ah, I found a comment in Compose's sources:
```
	// Starting from API version 1.44, the Engine will take several EndpointsConfigs
	// so we can pass all the extra networks we want the container to be connected to
	// in the network configuration instead of connecting the container to each extra
	// network individually after creation.
```
I guess that's the trick then... (For API version < 1.44, it simply passes on the first network if networks are provided for a service, or the default network.)

PR #940, opened 2024-07-25
Docker* connection plugins: allow to pass extra environment variables when running commands

### LINKED ISSUE #937: community.docker.docker add environment var overwrite
##### SUMMARY
Please add an option to specify environment variables. Sadly lthel existing `ansible_docker_extra_args` cannot be used to pass the `--env` option therefore a new additional option would be required.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
community.docker.docker
community.docker.docker_api

##### ADDITIONAL INFORMATION

```yaml
- name: GitLab | Add container to inventory
  ansible.builtin.add_host:
    host: '{{ gitlab_container_name }}'
    ansible_connection: docker
    ansible_docker_extra_args: '-H {{ gitlab_dockerhost }}'
    ansible_docker_extra_env:
      PATH: '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'
    ansible_python_interpreter: '/usr/bin/python3'
    ansible_shell_executable: '/bin/bash'
    ansible_become: false
```

felixfontein:
##### SUMMARY
Fixes #937.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
docker connection plugin
docker_api connection plugin

### DIFF
--- a/changelogs/fragments/940-connection-env.yml
+++ b/changelogs/fragments/940-connection-env.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "docker, docker_api connection plugins - allow to pass extra environment variables when executing commands with the new ``extra_env`` option (https://github.com/ansible-collections/community.docker/issues/937, https://github.com/ansible-collections/community.docker/pull/940)."
--- a/plugins/connection/docker.py
+++ b/plugins/connection/docker.py
@@ -232,7 +232,7 @@ class Connection(ConnectionBase):
                             'wrapped in quotes to avoid them being interpreted. {1}: {2!r}'
                             .format(what.lower(), what, val)
                         )
-                local_cmd += [b'-e', b'{0}={1}'.format(to_bytes(k, errors='surrogate_or_strict'), to_bytes(v, errors='surrogate_or_strict'))]
+                local_cmd += [b'-e', b'%s=%s' % (to_bytes(k, errors='surrogate_or_strict'), to_bytes(v, errors='surrogate_or_strict'))]
 
         # -i is needed to keep stdin open which allows pipelining to work
         local_cmd += [b'-i', self.get_option('remote_addr')] + cmd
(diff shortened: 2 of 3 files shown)


agowa:
I tested your pr locally and it appears to work just fine. However after looking into the code I'd like to bring your attention to the little (often overlooked) detail of yaml that the key of a dict can be any arbitrary node and even contain itself... 
> The content of a mapping node is an unordered set of key/value node pairs, with the restriction that each of the keys is [unique](https://yaml.org/spec/1.2.2/#node-comparison). YAML places no further restrictions on the nodes. In particular, keys may be arbitrary nodes, the same node may be used as the value of several key/value pairs and a mapping could even contain itself as a key or a value.

And a "node" in this context is defined as:
> A YAML node [represents](https://yaml.org/spec/1.2.2/#representation-graph) a single [native data structure](https://yaml.org/spec/1.2.2/#representing-native-data-structures). Such nodes have content of one of three kinds: scalar, sequence or mapping.

https://yaml.org/spec/1.2.2/#3211-nodes

I do not know how much the parser within ansible-core will already filter out, however as you did a type check on the value doing another one on the key may also be a good idea.

felixfontein:
> I do not know how much the parser within ansible-core will already filter out,

It does not filter anything out, and in fact you can even got worse things since you could create a lookup plugin or Jinja2 filter that returns arbitrary Python objects and pass them on to the plugin. So basically the dictionary can contain arbitrary Python objects as keys or values. (For modules this isn't possible, there everything is passed through JSON so dictionary keys can only be strings.)

> however as you did a type check on the value doing another one on the key may also be a good idea.

If you look more closely you can see that I actually check both the key *and* the value (`for val, what in ((k, 'Key'), (v, 'Value')):`, here `k, v` are key and value in the dictionary), exactly for this reason :)

agowa:
oh, sorry then. I must have overlooked that one line. I somehow directly jumped to the "isinstanceof" one...

PR #987, opened 2024-11-11
Add 'detach' option to docker_stack module

aliou-sidibe:
##### SUMMARY
 Add missing option --detach: to control immediate exit behavior on stack deployment/remove

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Docker stack

### DIFF
--- a/plugins/modules/docker_stack.py
+++ b/plugins/modules/docker_stack.py
@@ -57,6 +57,14 @@ options:
         current stack definition.
     type: bool
     default: false
+  detach:
+    description:
+      - If V(false), the C(--detach=false) option is added to the C(docker stack deploy) command,
+        allowing Docker to wait for tasks to converge before exiting.
+      - If V(true) (default), Docker exits immediately instead of waiting for tasks to converge.
+    type: bool
+    default: true
+    version_added: 4.1.0
   with_registry_auth:
     description:
       - If true will add the C(--with-registry-auth) option to the C(docker stack deploy) command.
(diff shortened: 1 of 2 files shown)


plugins/modules/docker_stack.py
```diff
@@ -57,6 +57,12 @@
         current stack definition.
     type: bool
     default: false
+  detach:
+    description:
+      - If false will add the C(--detach=false) option to the C(docker stack deploy) command.
```
felixfontein, line 62:
```suggestion
      - If V(false) will add the C(--detach=false) option to the C(docker stack deploy) command.
```

plugins/modules/docker_stack.py
```diff
@@ -57,6 +57,12 @@
...
+        This will have docker exit immediately instead of waiting for the stack services to converge.
```
felixfontein, line 63: The way I understand the description of `--detach`, it sounds that `--detach=true` (the default) is "Exit immediately instead of waiting for the stack services to converge". Are you sure that `--detach=false` means that Docker exits immediately?

felixfontein (commented):
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks!

plugins/modules/docker_stack.py
aliou-sidibe: Setting `--detach=false` means Docker waits for all tasks to converge before exiting. This implies that Docker does not detach and will block until the services reach a `running` state or are `healthy` if a health check is configured. This ensures that the deployment is stable before the command completes.

On the other hand, `--detach=true` (the default) means Docker exits immediately after starting the services, without waiting for them to converge.

So, `--detach=false` is indeed what makes Docker wait for the tasks to converge before exiting.

felixfontein: Ok, in that case you need to update the description. Right now it says the opposite :)

felixfontein (approved):

felixfontein:
@aliou-sidibe thanks for your contribution!

Issue #891, opened 2024-06-23
podman-docker incompatibilities in `common_cli.py`: `KeyError: 'ApiVersion'`


pavelzw:
##### SUMMARY

The plugin fails when using `podman-docker` because it expects a specific format output format in `docker version --format '{{ json . }}'` which is not supported by `podman-docker:

https://github.com/ansible-collections/community.docker/blob/eddeb916973dbbd82298132bed285f803bd505d1/plugins/module_utils/common_cli.py#L83

```json
$ docker version --format '{{ json . }}' | jq
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
{
  "Client": {
    "APIVersion": "5.1.1",
    "Version": "5.1.1",
    "GoVersion": "go1.22.3",
    "GitCommit": "",
    "BuiltTime": "Tue Jun  4 02:00:00 2024",
    "Built": 1717459200,
    "OsArch": "linux/amd64",
    "Os": "linux"
  }
}
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

community.docker/plugins/module_utils/common_cli.py

##### ANSIBLE VERSION

```paste below
ansible [core 2.16.6]
  config file = None
  configured module search path = ['/Users/pavel/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = .../.pixi/envs/default/lib/python3.12/site-packages/ansible
  ansible collection location = /Users/pavel/.ansible/collections:/usr/share/ansible/collections
  executable location = .../.pixi/envs/default/bin/ansible
  python version = 3.12.3 | packaged by conda-forge | (main, Apr 15 2024, 18:35:20) [Clang 16.0.6 ] (.../.pixi/envs/default/bin/python3.12)
  jinja version = 3.1.4
  libyaml = True
```

##### COLLECTION VERSION

```paste below
❯ ansible-galaxy collection list community.docker

# /Users/pavel/.ansible/collections/ansible_collections
Collection       Version
---------------- -------
community.docker 3.10.4 

# .../.pixi/envs/default/lib/python3.12/site-packages/ansible_collections
Collection       Version
---------------- -------
community.docker 3.9.0  
```

##### CONFIGURATION

```paste below
CONFIG_FILE() = None
```

##### OS / ENVIRONMENT

macOS M1

target OS: fedora 40 with `podman` and `podman-docker` installed

```
$ dnf list installed | grep podman
[...]

PR #1021, opened 2024-12-28
Determine the compose version via a CLI call and not the docker API.

apollo13:
##### SUMMARY
Call the CLI to determine the compose version before talking to the API.
Refs #891

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
docker_compose_v2

### DIFF
--- a/changelogs/fragments/1021-docker_compose_v2-version-detection.yml
+++ b/changelogs/fragments/1021-docker_compose_v2-version-detection.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "docker_compose_v2* modules - determine compose version with ``docker compose version`` and only then fall back to ``docker info`` (https://github.com/ansible-collections/community.docker/pull/1021)."
--- a/plugins/module_utils/compose_v2.py
+++ b/plugins/module_utils/compose_v2.py
@@ -735,7 +735,10 @@ class BaseComposeManager(DockerBaseClass):
         if rc:
             return None
         try:
-            return json.loads(version_info)['version'].lstrip('v')
+            version = json.loads(version_info)['version']
+            if version == 'dev':
+                return None
+            return version.lstrip('v')
         except Exception:
             return None
 
--- a/plugins/module_utils/compose_v2.py
+++ b/plugins/module_utils/compose_v2.py
@@ -701,16 +701,7 @@ class BaseComposeManager(DockerBaseClass):
         self.env_files = parameters['env_files']
         self.profiles = parameters['profiles']
 
-        compose = self.client.get_client_plugin_info('compose')
-        if compose is None:
-            self.fail('Docker CLI {0} does not have the compose plugin installed'.format(self.client.get_cli()))
-        if compose['Version'] == 'dev':
-            self.fail(
-                'Docker CLI {0} has a compose plugin installed, but it reports version "dev".'
-                ' Please use a version of the plugin that returns a proper version.'
-                .format(self.client.get_cli())
-            )
-        compose_version = compose['Version'].lstrip('v')
+        compose_version = self.get_compose_version()
         self.compose_version = LooseVersion(compose_version)
         if self.compose_version < LooseVersion(min_version):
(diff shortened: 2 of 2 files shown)


plugins/module_utils/compose_v2.py
```diff
@@ -701,23 +701,30 @@ def __init__(self, client, min_version=MINIMUM_COMPOSE_VERSION):
...
-        compose_version = compose['Version'].lstrip('v')
-        self.compose_version = LooseVersion(compose_version)
-        if self.compose_version < LooseVersion(min_version):
-            self.fail('Docker CLI {cli} has the compose plugin with version {version}; need version {min_version} or later'.format(
-                cli=self.client.get_cli(),
-                version=compose_version,
-                min_version=min_version,
-            ))
+        # Check "docker compose version" first as this allows for "podman compose" to work as well
+        rc, version, stderr = self.client.call_cli('compose', 'version', '--short')
```
felixfontein, line 705: I think it would be better to use the JSON output and parse that.
```suggestion
        rc, version, stderr = self.client.call_cli('compose', 'version', '--format', 'json')
```

felixfontein (commented):
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks. (I think this is more like a feature than a bugfix though. The module was never supposed to work with Podman, so it's not a bug that it didn't work with it.)

Regarding the implementation: I think it would be a lot cleaner if you could move the code into multiple methods. One method `get_docker_compose_version()` (or similar) that calls `docker compose version --formta json`and returns the version as a string, or `None` if it cannot detemrine the version; and another method `get_compose_plugin_version()` (or similar) that contains the current code (and returns the version as a string, or fails). And maybe a third method which calls the other two and always returns the version as a string.

apollo13:
> Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks. (I think this is more like a feature than a bugfix though. The module was never supposed to work with Podman, so it's not a bug that it didn't work with it.)

Sure thing, I wanted to know whether it passes CI first :) Added it as minor_change.

> Regarding the implementation: I think it would be a lot cleaner if you could move the code into multiple methods.

Done I think, let's see what CI says, locally it did work, but my test env is "limited" to a single python and docker.

changelogs/fragments/1021-docker_compose_v2-version-detection.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "docker_compose_v2 - determine compose version via CLI and fall back to the API (https://github.com/ansible-collections/community.docker/pull/1021).
```
felixfontein, line 2:
```suggestion
  - "docker_compose_v2* modules - determine compose version with ``docker compoes version`` and only then fall back to ``docker info`` (https://github.com/ansible-collections/community.docker/pull/1021)."
```

felixfontein (commented):
Looks good! I've just have suggestion for the changelog fragment. This is about all docker_compose_v2* modules, not just docker_compose_v2 itself.

plugins/module_utils/compose_v2.py
```diff
@@ -736,6 +727,30 @@ def __init__(self, client, min_version=MINIMUM_COMPOSE_VERSION):
...
 
+    def get_compose_version(self):
+        return self.get_compose_version_from_cli() or self.get_compose_version_from_api()
+
+    def get_compose_version_from_cli(self):
+        rc, version_info, stderr = self.client.call_cli('compose', 'version', '--format', 'json')
+        if rc:
+            return None
+        try:
+            return json.loads(version_info)['version'].lstrip('v')
```
felixfontein, line 738: One final thing, let's handle `dev` here as well:
```suggestion
            version = json.loads(version_info)['version']
            if version == 'dev':
                return None
            return version.lstrip('v')
```

felixfontein (approved):

apollo13:
Thank you for the quick turnaround and support (as always ;))

PR #969, opened 2024-09-22
Add Docker Compose v2 exec and run modules

felixfontein:
##### SUMMARY
Ref: #956

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
docker_compose_v2_exec
docker_compose_v2_run

PR #1034, opened 2025-01-31
fix: docker_compose_v2_run: labels

Tardo:
##### SUMMARY
Fixes the support for labels. Without these changes, the module has a conflict between the type of parameter it expects and the one it tries to sanitize.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
docker_compose_v2_run

### DIFF
--- a/changelogs/fragments/1034-do-not-sanitize-labels.yaml
+++ b/changelogs/fragments/1034-do-not-sanitize-labels.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - docker_compose_v2_run - the module has a conflict between the type of parameter it expects and the one it tries to sanitize. Fix removes the label sanitization step because they are already validated by the parameter definition (https://github.com/ansible-collections/community.docker/pull/1034).
--- a/plugins/modules/docker_compose_v2_run.py
+++ b/plugins/modules/docker_compose_v2_run.py
@@ -253,8 +253,6 @@ from ansible_collections.community.docker.plugins.module_utils.compose_v2 import
     common_compose_argspec_ex,
 )
 
-from ansible_collections.community.docker.plugins.module_utils.util import sanitize_labels
-
 
 class ExecManager(BaseComposeManager):
     def __init__(self, client):
@@ -424,7 +422,6 @@ def main():
         needs_api_version=False,
         **argspec_ex
     )
-    sanitize_labels(client.module.params['labels'], 'labels', client)
 
     try:
         manager = ExecManager(client)


plugins/modules/docker_compose_v2_run.py
```diff
@@ -253,7 +252,10 @@
     common_compose_argspec_ex,
 )
 
-from ansible_collections.community.docker.plugins.module_utils.util import sanitize_labels
+from ansible_collections.community.docker.plugins.module_utils.util import (
+    sanitize_labels,
```
felixfontein, line 256: The correct way to fix this is to create a new function `sanitize_list()` in `ansible_collections.community.docker.plugins.module_utils.util` with a similar interface (and implementation) to `sanitize_labels()`, except that it processes a list and not dictionary values.

(And I'm really looking forward to getting rid of Python 2.7, so we can add typing hints to prevent such bugs...)

felixfontein (changes requested):
Thanks for your contribution!

plugins/modules/docker_compose_v2_run.py
Tardo: Thank you for the hints. It should be done now.

plugins/module_utils/util.py
```diff
@@ -301,6 +301,28 @@ def fail(msg):
...
+            module.fail_json(msg=msg)
+        raise ValueError(msg)
+
+    if labels is None:
+        return
+    for index, label in enumerate(labels):
+        if not isinstance(label, string_types):
+            fail(
+                "The item {item!r} of {field} is not a string!".format(
+                    field=labels_field, item=label))
```
felixfontein, line 318: In the other function this is checking the key, not the value. This isn't needed here.
```suggestion
```

plugins/modules/docker_compose_v2_run.py
```diff
@@ -424,7 +424,7 @@ def main():
         needs_api_version=False,
         **argspec_ex
     )
-    sanitize_labels(client.module.params['labels'], 'labels', client)
+    sanitize_labels_list(client.module.params['labels'], 'labels', client)
```
felixfontein, line 427: Thinking a bit more about this, I think we can simply remove this line here and be done with it. Ansible already ensured (due to `type=list` and `elements=str`) that all elements of `client.module.params['labels']` are strings.

Sorry for not noticing that earlier!

Tardo: Yes, I thought the same thing when I saw it.... I imagined that maybe it was left because of some kind of “redundancy”.

It still hurts me that in all the other components it is a dictionary and here, by whims of fate, it is a list of strings.

felixfontein: It was an oversight by me when adding the module (#969). Unfortunately nobody commented on it so far (especially not before merging the PR).

Tardo: Oook! changes done :)

felixfontein:
Thanks! Can you add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)?

Tardo:
Done! I hope everything is correct! 🤞🏻 :)

changelogs/fragments/1034-do-not-sanitize-labels.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - docker_compose_v2_run - the component has a conflict between the type of parameter it expects and the one it tries to sanitize. Fix removes the label sanitization step because they are already validated by the parameter definition. (https://github.com/ansible-collections/community.docker/pull/1034).
```
felixfontein, line 2:
```suggestion
  - docker_compose_v2_run - the component has a conflict between the type of parameter it expects and the one it tries to sanitize. Fix removes the label sanitization step because they are already validated by the parameter definition (https://github.com/ansible-collections/community.docker/pull/1034).
```

Tardo: Done!

changelogs/fragments/1034-do-not-sanitize-labels.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - docker_compose_v2_run - the module has a conflict between the type of parameter it expects and the one it tries to sanitize. Fix removes the label sanitization step because they are already validated by the parameter definition. (https://github.com/ansible-collections/community.docker/pull/1034).
```
felixfontein, line 2: There's one thing you missed :)
```suggestion
  - docker_compose_v2_run - the module has a conflict between the type of parameter it expects and the one it tries to sanitize. Fix removes the label sanitization step because they are already validated by the parameter definition (https://github.com/ansible-collections/community.docker/pull/1034).
```

Tardo: 😭 xD

felixfontein (approved):

felixfontein:
@Tardo thanks a lot for fixing this!

PR #1039, opened 2025-02-03
Add docker_context_info module

felixfontein:
##### SUMMARY
The goal is to have a module to query information on Docker contexts.

This also vendors the contexts part of the Docker SDK for Python (https://github.com/docker/docker-py/tree/main/docker/context).

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
docker_context_info


felixfontein:
@ssbarnea this might be useful for you when trying to mix community.docker with Docker CLI tools (which consider Docker contexts). Something like
```yaml

- name: Get current context
  community.docker.docker_context_info:
    only_current: true
  register: docker_current_context
- name: Run community.docker modules with current context
  module_defaults:
    group/community.docker.docker: "{{ docker_current_context.contexts[0].config }}"
  block:
    - name: Task using the current context
        image: ubuntu:latest
        name: ubuntu
        state: started
```
will allow you to run community.docker modules with Docker's current context.

plugins/module_utils/_api/context/api.py
```diff
@@ -0,0 +1,242 @@
+# -*- coding: utf-8 -*-
+# This code is part of the Ansible collection community.docker, but is an independent component.
+# This particular file, and this file only, is based on the Docker SDK for Python (https://github.com/docker/docker-py/)
+#
+# Copyright (c) 2016-2025 Docker, Inc.
```
markuman, line 5: Why is this copyright be docker? is it copied from some upstream sdk?

plugins/module_utils/_api/context/config.py
```diff
@@ -0,0 +1,103 @@
...
+    return os.path.join(context_dir, "tls")
+
+
+def get_context_host(path=None, tls=False):
+    host = parse_host(path, IS_WINDOWS_PLATFORM, tls)
+    if host == DEFAULT_UNIX_SOCKET:
+        # remove http+ from default docker socket url
+        if host.startswith("http+"):
+            host = host[5:]
+    return host
```
markuman, line 103:
```suggestion
    host = parse_host(path, IS_WINDOWS_PLATFORM, tls)
    
    # remove http+ from default docker socket url if exists
    return host.split("http+")[-1]
```

But I don't know how the host looks like when it's not a `DEFAULT_UNIX_SOCKET`.

plugins/module_utils/_api/context/api.py
felixfontein: Yep, it's copied from https://github.com/docker/docker-py/

plugins/module_utils/_api/context/config.py
felixfontein: I'd prefer not to change this, since it will be some divergence between the original Docker SDK for Python and this vendored version of the code that might break at some point in funny/unexpected ways.

Docker SDK for Python uses `http+` or `https+` before `unix://` to indicate the exact protocol to use (with or without TLS), and at the same time tries to stay compatible with the Docker CLI programs that only use `unix://` without `http+` or `https+` (and don't understand that). This check here is one of the places where it tries to return the same value that the CLI tools do, *at least in one very specific case* (namely the default Docker socket).

markuman (approved):

felixfontein:
@markuman thanks a lot for reviewing this!

PR #1045, opened 2025-02-21
add `--yes` parameter to `docker compose up`

londondaintta:
##### SUMMARY

Add a variable for the `docker compose up` parameter `yes` to assume "yes" as answer to all prompts and run non-interactively. This prevents issues where the module hangs if a volume spec changes in the compose file.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

docker_compose_v2

##### ADDITIONAL INFORMATION

```yaml
    - name: Start compose environment with yes
      community.docker.docker_compose_v2:
        project_src: compose
        assume_yes: true
        state: present
```

### DIFF
--- a/changelogs/fragments/1045-docker-compose-up-yes.yaml
+++ b/changelogs/fragments/1045-docker-compose-up-yes.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - docker_compose_v2 - add ``yes`` parameter for ``docker compose up`` (https://github.com/ansible-collections/community.docker/pull/1045).
+  - docker_compose_v2 - add ``assume_yes`` parameter for ``docker compose up`` (https://github.com/ansible-collections/community.docker/pull/1045).
--- a/changelogs/fragments/1045-docker-compose-up-yes.yaml
+++ b/changelogs/fragments/1045-docker-compose-up-yes.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - docker_compose_v2 - add ``yes`` parameter for ``docker compose up`` (https://github.com/ansible-collections/community.docker/pull/1045).
--- a/plugins/modules/docker_compose_v2.py
+++ b/plugins/modules/docker_compose_v2.py
@@ -163,7 +163,7 @@ options:
   yes:
     description:
       - When O(yes=true), pass C(--yes) to assume "yes" as answer to all prompts and run non-interactively.
-    type: int
+    type: bool
     version_added: TODO
 
 author:
--- a/plugins/modules/docker_compose_v2.py
+++ b/plugins/modules/docker_compose_v2.py
@@ -160,6 +160,11 @@ options:
       - When O(wait=true), wait at most this amount of seconds.
     type: int
     version_added: 3.8.0
+  yes:
+    description:
+      - When O(yes=true), pass C(--yes) to assume "yes" as answer to all prompts and run non-interactively.
+    type: int
+    version_added: TODO
 
 author:
   - Felix Fontein (@felixfontein)
@@ -465,6 +470,7 @@ class ServicesManager(BaseComposeManager):
         self.scale = parameters['scale'] or {}
         self.wait = parameters['wait']
         self.wait_timeout = parameters['wait_timeout']
+        self.yes = parameters['yes']
 
         for key, value in self.scale.items():
             if not isinstance(key, string_types):
@@ -522,6 +528,8 @@ class ServicesManager(BaseComposeManager):
             args.append('--no-start')
         if dry_run:
             args.append('--dry-run')
+        if self.yes:
+            args.append('--yes')
         args.append('--')
         for service in self.services:
             args.append(service)
@@ -656,6 +664,7 @@ def main():
         wait=dict(type='bool', default=False),
         wait_timeout=dict(type='int'),
         ignore_build_events=dict(type='bool', default=True),
+        yes=dict(type='bool', default=False),
     )
     argspec_ex = common_compose_argspec_ex()
     argument_spec.update(argspec_ex.pop('argspec'))
--- a/plugins/modules/docker_compose_v2.py
+++ b/plugins/modules/docker_compose_v2.py
@@ -453,6 +453,8 @@ from ansible_collections.community.docker.plugins.module_utils.compose_v2 import
     is_failed,
 )
 
+from ansible_collections.community.docker.plugins.module_utils.version import LooseVersion
+
 
 class ServicesManager(BaseComposeManager):
     def __init__(self, client):
@@ -474,7 +476,8 @@ class ServicesManager(BaseComposeManager):
         self.scale = parameters['scale'] or {}
         self.wait = parameters['wait']
         self.wait_timeout = parameters['wait_timeout']
-        self.yes = parameters['assume_yes']
+        if self.compose_version >= LooseVersion('2.32.0'):
+          self.yes = parameters['assume_yes']
 
         for key, value in self.scale.items():
             if not isinstance(key, string_types):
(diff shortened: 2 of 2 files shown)


plugins/modules/docker_compose_v2.py
```diff
@@ -160,6 +160,12 @@
       - When O(wait=true), wait at most this amount of seconds.
     type: int
     version_added: 3.8.0
+  yes:
+    description:
+      - When O(yes=true), pass C(--yes) to assume "yes" as answer to all prompts and run non-interactively.
+    type: bool
+    default: false
+    version_added: TODO
```
felixfontein, line 168:
```suggestion
    version_added: 4.5.0
```

plugins/modules/docker_compose_v2.py
```diff
@@ -160,6 +160,12 @@
       - When O(wait=true), wait at most this amount of seconds.
     type: int
     version_added: 3.8.0
+  yes:
+    description:
+      - When O(yes=true), pass C(--yes) to assume "yes" as answer to all prompts and run non-interactively.
```
felixfontein, line 165:
```suggestion
      - When O(yes=true), pass C(--yes) to assume "yes" as answer to all prompts and run non-interactively.
      - Right now a prompt is asked whenever a non-matching volume should be re-created. O(yes=false)
        results in the question being answered by "no", which will simply re-use the existing volume.
```

plugins/modules/docker_compose_v2.py
```diff
@@ -160,6 +160,12 @@
       - When O(wait=true), wait at most this amount of seconds.
     type: int
     version_added: 3.8.0
+  yes:
```
felixfontein, line 163: I'm not sure `yes` is a good option name. One problem is that "yes" is a YAML 1.1 boolean value, so it will require quoting the option to be able to use it (unless Ansible does some magic).

How about `approve_questions` or something like that?

felixfontein (commented):
Thanks for your contribution!

The integration test failures are unrelated, the nightly tests were failing as well. (This started yesterday, after the release of Docker Compose 2.33.0.)

The `--yes` option was added in Compose 2.32.0 (https://github.com/docker/compose/pull/12363).

felixfontein:
Could you also add a check to `__init__()` that `self.compose_version >= LooseVersion('2.32.0')` when the option is set to `true` (you'll need `from ansible_collections.community.docker.plugins.module_utils.version import LooseVersion`)?

felixfontein:
> The integration test failures are unrelated, the nightly tests were failing as well. (This started yesterday, after the release of Docker Compose 2.33.0.)

The problem is actually Docker CLI 28.0.0, which generates some unexpected stderr output. This has been fixed in https://github.com/docker/cli/pull/5854, I hope there will be a bugfix release soon.

plugins/modules/docker_compose_v2.py
londondaintta: Yes, that's a good point. Perhaps `assume_yes` is a little more obvious?

plugins/modules/docker_compose_v2.py
londondaintta: `results in the question being answered by "no"` is not quite right. It actually causes the command to hang as it waits for stdin.

felixfontein: Did you test this? From Docker Compose's code I would expect it to not be able to read anything from stdin and assume "no".

changelogs/fragments/1045-docker-compose-up-yes.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - docker_compose_v2 - add ``yes`` parameter for ``docker compose up`` (https://github.com/ansible-collections/community.docker/pull/1045).
```
felixfontein, line 2:
```suggestion
  - docker_compose_v2 - add ``assume_yes`` parameter for ``docker compose up`` (https://github.com/ansible-collections/community.docker/pull/1045).
```

plugins/modules/docker_compose_v2.py
```diff
@@ -465,6 +476,10 @@ def __init__(self, client):
         self.scale = parameters['scale'] or {}
         self.wait = parameters['wait']
         self.wait_timeout = parameters['wait_timeout']
+        if self.compose_version >= LooseVersion('2.32.0'):
+          self.yes = parameters['assume_yes']
+        else:
+          self.yes = False
```
felixfontein, line 482: I would fail if `assume_yes=true` is supplied with a too old version, instead of silently ignoring it:
```suggestion
        self.yes = parameters['assume_yes']
        if self.compose_version >= LooseVersion('2.32.0') and self.yes:
            self.fail('assume_yes=true needs Docker Compose 2.32.0 or newer, not version %s' % (self.compose_version, ))
```

plugins/modules/docker_compose_v2.py
londondaintta: Yes, it will prompt "Volume X exists but doesn't match configuration in compose file. Recreate (data will be lost)?" then read from stdin. 

https://github.com/docker/compose/pull/12363/files#diff-1f4d596600c620b3df55dbfbc63e50a6f43b356865fae411c553d0937fbb6d91R1468

felixfontein: Which part of my post does your "yes" mean? :)

https://github.com/ndeloof/compose/blob/4e7b6dfc3d0f725bc561cfe88df68833b5863811/pkg/prompt/prompt.go#L99 should try to read from stdin and notice that there's nothing to read, whence `answer` will be the empty string. `utils.StringToBool("")` should end up as `false` (https://github.com/ndeloof/compose/blob/4e7b6dfc3d0f725bc561cfe88df68833b5863811/pkg/utils/stringutils.go#L40; unfortunately `strconv.ParseBool()`'s docs doesn't say which value is returned in case of an error, but from experiments it seems to be `false`, and the implementation concurs: https://github.com/golang/go/blob/master/src/strconv/atob.go#L17).

I haven't checked out whether this really happens, just pieced it together from other programs I've run with `module.run_command()` and the above.

londondaintta: Sorry, yes as in yes, I have tested it.

Here's a fairly simple way to reproduce:

```bash
mkdir /tmp/{first,second}

cat << EOF > compose.yaml
services:
  client:
    image: busybox
    command: "tail -f /dev/null"
    volumes:
      - client_in:/in

volumes:
  client_in:
    driver: local
    driver_opts:
      type: "none"
      o: "bind"
      device: "/tmp/first"
EOF

docker compose up -d

sed -i 's/first/second/g' foo

docker compose up -d
# see the prompt wait
```

londondaintta: So in my play I set `recreate: always` which then causes this to hang.

felixfontein: I cannot reproduce it hanging with `recreate: always`. For me, the task then takes ~12 seconds, where ~10 seconds is the default timeout for stopping a container.

londondaintta: I will try and see if I can recreate this more reliably, in the mean time is this ok to be merged?

felixfontein: Sure. I'll merge it later today; either by then Arch updated their Compose package and CI will pass, or I'll just merge anyway if it takes too long...

felixfontein (approved):

felixfontein:
@londondaintta thanks for your contribution!

jonasgeiler:
Uhm @londondaintta @felixfontein just a note: There is no `--yes` flag. Only a `--y` flag.
Try running `docker compose up --yes`. It will fail.

I don't know why the Docker maintainers didn't correctly name this flag...

felixfontein:
Hmm, at least they were consequential, since there are three subcommands that all use `-y` and `--y` :facepalm:

PR #1046, opened 2025-02-21
[DO NOT MERGE] Debug compose_v2 CI

felixfontein:
##### SUMMARY
Since Docker Compose 2.33.0 CI is failing. (Maybe it's fixed by 2.33.1, which has been released a couple of hours ago.)

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/plugins/module_utils/compose_v2.py
+++ b/plugins/module_utils/compose_v2.py
@@ -7,6 +7,8 @@ from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
 
+# force tests
+
 import json
 import os
 import re


felixfontein:
While Compose 2.33.1 is out, the faililng tests still run with 2.33.0. Note that some *passing* tests also run with 2.33.0, for example the Arch Linux tests. My guess is that Arch builds Compose slightly different than Docker themselves.

It's too bad that Docker themselves are not providing Debian/Ubuntu packages for Compose 2.33.1 yet...

felixfontein:
Looks like this is actually a bug in Docker CLI, not in Docker Compose. It was introduced in https://github.com/docker/cli/pull/5229 (which got included in 28.0.0), and fixed in https://github.com/docker/cli/pull/5854, which unfortunately isn't included in 28.0.0 yet.

felixfontein:
28.0.1 has been released, and tests now pass, except on Arch Linux, which upgraded its Docker to 28.0.0 by now, but not to 28.0.1 ;-)

(Arch Linux isn't failling here since I only restarted the failing tests, and the previous successful run was several days ago when Arch still had < 28.0.0.)

PR #1054, opened 2025-03-10
fix: use correct flag for `assume_yes`

jonasgeiler:
##### SUMMARY

The correct flag is `--y`, not `--yes`, as mentioned in my comment on #1045.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

docker_compose_v2

##### ADDITIONAL INFORMATION

For some reason the flag is called `--y`, not `--yes` as is usual. I think this might be a mistake from the Docker maintainers, but for now this is a bug and causes an error if `--yes` is used.

```paste below
# (Not sure I understand what belong in here)
```

### DIFF
--- a/plugins/modules/docker_compose_v2.py
+++ b/plugins/modules/docker_compose_v2.py
@@ -162,7 +162,7 @@ options:
     version_added: 3.8.0
   assume_yes:
     description:
-      - When O(assume_yes=true), pass C(--yes) to assume "yes" as answer to all prompts and run non-interactively.
+      - When O(assume_yes=true), pass C(--y) to assume "yes" as answer to all prompts and run non-interactively.
       - Right now a prompt is asked whenever a non-matching volume should be re-created. O(assume_yes=false)
         results in the question being answered by "no", which will simply re-use the existing volume.
       - This option is only available on Docker Compose 2.32.0 or newer.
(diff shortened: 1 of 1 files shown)


plugins/modules/docker_compose_v2.py
```diff
@@ -537,7 +537,7 @@ def get_up_cmd(self, dry_run, no_start=False):
         if dry_run:
             args.append('--dry-run')
         if self.yes:
-            args.append('--yes')
+            args.append('-y')
```
felixfontein, line 540: I would add a comment here:
```suggestion
            # Note that the long form is '--y', and not '--yes' as one would expect. Since this looks like a bug we're using the short form.
            args.append('-y')
```

felixfontein (commented):
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks!

jonasgeiler:
> @felixfontein wrote: Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks!

Done! Hope I did it right.

felixfontein (approved):

felixfontein:
@jonasgeiler thanks for your contribution!

jonasgeiler:
@felixfontein Just so you know Docker Compose Plugin [v2.34.0](https://github.com/docker/compose/releases/tag/v2.34.0) fixed the flag. It's now `--yes` as it is supposed to be.

PR #1059, opened 2025-03-17
fix(docker_compose_v2): fix version check

londondaintta:
##### SUMMARY

The version check added in #1045 for `assume_yes` was the wrong way way round.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

docker_compose_v2


londondaintta:
Only got around to testing this fix and noticed the version check was wrong! @felixfontein is it worth adding a simple test?

felixfontein:
Thanks for fixing this! No, I don't think it needs a test.

felixfontein:
@londondaintta thanks!

PR #1071, opened 2025-04-26
Add YAML linting to CI, and reformat YAML files and content

felixfontein:
##### SUMMARY
Uses antsibull-nox to run yamllint both for YAML files and YAML content in plugin docs.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
Hmm, the problem is that passing octal numbers as a string to `docker_container_copy_into`'s `mode` parameter does not work. Since its type is `int`, ansible-core will parse `"0777"` as the decimal integer 777 (which is 01411 in octal).

We can change the type of `mode` to `raw` to handle this better (similar to `mode` is handled by ansible-core's file attribute handling), but that's potentially breaking users.

Maybe the best way is to make it `raw`, and provide parameter which allows to define the behavior (and soon deprecate the current default, and warn if something happens due to it).

PR #1081, opened 2025-05-21
Add RHEL 10.0 to CI

felixfontein:
##### SUMMARY
Ref: https://github.com/ansible/ansible/pull/85189
Ref: https://forum.ansible.com/t/ansible-test-devel-rhel-10-0-added/42892

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -198,6 +198,8 @@ stages:
         parameters:
           testFormat: devel/{0}
           targets:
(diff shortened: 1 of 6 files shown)


felixfontein:
Looks like Docker doesn't work well with RHEL 10 since it still uses iptables, and something isn't there it expects...

felixfontein:
Ref: https://github.com/docker/for-linux/issues/1537

Issue #956, opened 2024-08-16
Add more Docker Compose modules


felixfontein:
##### SUMMARY
The following *could* make sense as new modules. See https://github.com/ansible-collections/community.docker/issues/941#issuecomment-2251219797 for more details.

- [ ] `build`: makes sense as its own module docker_compose_v2_build
- [ ] `cp`: makes sense as its own module, or two (copy into/out of container?)
- [ ] `exec`: makes sense as its own module docker_compose_v2_exec
- [ ] `push`: makes sense as its own module docker_compose_v2_push
- [ ] `run`: makes sense as its own module docker_compose_v2_run
- [ ] `wait`: *might* make sense as its own module
- [ ] docker_compose_v2_info

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
new modules

lnlyssg:
I ended up here as I'm looking to create a playbook to update my Mastodon instance. As per the linked comment I can see you are looking for any relevant use cases before adding any of these so I'd like to request `run` please. The Mastodon update instructions sometimes require running commands such as `docker-compose run --rm web bundle exec rails db:migrate` (and/or others) which I would have to run via the shell plugin currently.

felixfontein:
community.docker 3.13.0 (released earlier today) contains two new modules `docker_compose_v2_exec` and `docker_compose_v2_run`.

Issue #941, opened 2024-07-25
docker_compose_v2_pull: add `ignore_buildable` option (and/or `extra_args`)


mjbogusz:
##### SUMMARY

As the title says - it'd be great if the module had an equivalent of `--ignore-buildable` flag.

This would allow for a more smallest-step-possible approach, which I'm trying to achieve (`pull`, `build`, `up` all in separate steps).

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`docker_compose_v2_pull` module

##### ADDITIONAL INFORMATION

I believe a `ignore_buildable: bool` flag could be added, defaulting to `False`, which if set to `True` would add the `--ignore-buildable` flag to the `docker compose` call.

Alternatively (or additionally, as it's not exclusive) an `extra_args` flag would allow for better flexibility at the cost of slightly higher fragility of users' setups if used.

Example playbook (partial) showing what I'd like to achieve:
```yaml
tasks: 
  - name: Pull external docker images
    community.docker.docker_compose_v2_pull:
      project_src: /path/to/project
      # This is what I'd love here
      ignore_buildable: true
      # Alternatively, this would be great too
      extra_args: '--ignore-buildable'
      
  - name: Build internal docker images
    community.docker.docker_compose_v2:
      project_src: /path/to/project
      state: absent
      build: always

  - name: Start the thing
    community.docker.docker_compose_v2:
      project_src: /path/to/project
      state: present
      wait: true
```

felixfontein:
Options like `extra_args` are dangerous, I won't add anything like that. But `ignore_buildable` sounds good to me!

felixfontein:
#942 adds this option (and two more, while being at it...).

mjbogusz:
Nice, thanks!

That's what I thought as well about `extra_args` - it sounds great until everyone abuses it and then it breaks ;)

---

Also, for anyone trying to achieve what I've described in the first post: unfortunately the `build` part of my example playbook is not 100% correct either, as if I understand correctly it tries to do something like `docker compose down --build` which makes no sense.

For now I've worked around it by just throwing in a `ansible.builtin.command`, but maybe a `build_only` flag would be a way to go - I'll try digging more into it and potentially I'll post another feature request.

For a brief moment I've thought about a `docker_compose_v2_build` module but that would imply a need for a dozen others ;)

felixfontein:
[...]

PR #1090, opened 2025-06-17
Add docker_compose_v2_build module (#956)

mjbogusz:
##### SUMMARY
As the title states - add docker_compose_v2_build module.

Ref #956 and discussion from #941

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
docker_compose_v2_build

##### ADDITIONAL INFORMATION
For now this is a draft PR with only the module itself (so e.g. any tests are missing). I've only spent a bit of time digging into the existing codebase so there might be some obvious problems/errors in my code - feel free to point them out. If this looks OK and there's some interest I'll try to find time to flesh out the missing pieces.

The module's code is heavily based upon `_pull` equivalent, thus I've included @felixfontein in copyrights.

### DIFF
--- a/plugins/modules/docker_compose_v2_build.py
+++ b/plugins/modules/docker_compose_v2_build.py
@@ -1,0 +1,190 @@
+#!/usr/bin/python
+#
+# Copyright (c) 2023, Felix Fontein <felix@fontein.de>
+# Copyright (c) 2025, Maciej Bogusz (@mjbogusz)
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r"""
+module: docker_compose_v2_build
+
+short_description: Build a Docker compose project
+
+version_added: 4.7.0
+
+description:
+  - Uses Docker Compose to build images for a project.
+extends_documentation_fragment:
+  - community.docker.compose_v2
+  - community.docker.compose_v2.minimum_version
+  - community.docker.docker.cli_documentation
+  - community.docker.attributes
+  - community.docker.attributes.actiongroup_docker
+
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+  idempotent:
+    support: full
+
+options:
+  no_cache:
+    description:
+      - If set to V(true), will not use cache when building the images.
+    type: bool
+    default: false
+  pull:
+    description:
+      - If set to V(true), will attempt to pull newer version of the image
+    type: bool
+    default: false
+  with_dependencies:
+    description:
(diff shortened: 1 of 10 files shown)


plugins/modules/docker_compose_v2_build.py
```diff
@@ -0,0 +1,190 @@
...
+
+options:
+  no_cache:
+    description:
+      - If set to V(true), will not use cache when building the images.
+    type: bool
+    default: false
+  pull:
+    description:
+      - If set to V(true), will attempt to pull newer version of the image
```
felixfontein, line 44:
```suggestion
      - If set to V(true), will attempt to pull newer version of the image.
```

plugins/modules/docker_compose_v2_build.py
```diff
@@ -0,0 +1,190 @@
...
+    default: false
+  with_dependencies:
+    description:
+      - If set to V(true), also build services that are declared as dependencies.
+      - This only makes sense if O(services) is used.
+    type: bool
+    default: false
+  memory_limit:
+    description:
+      - Memory limit for the build container, in bytes. Not supported by BuildKit
```
felixfontein, line 55:
```suggestion
      - Memory limit for the build container, in bytes. Not supported by BuildKit.
```

felixfontein (commented):
Thanks for your contribution! From a first glance, this looks good!

mjbogusz:
I've rebased my branch, applied the suggestions, added references to this module in other parts of documentation and added some preliminary unit tests.

Tests are based on the ones for the docker_compose_v2_pull module, are not exhaustive and probably not even correct as-is, as their setup is a whole rabbit-hole I only had some time to look at the surface of - hints, suggestions or even pushes to my branch are very welcome.

felixfontein:
A big problem seems to be that it is impossible to figure out from the output whether the image is actually built (the actions always say "Built" and that's it, which usually means the image was already built before). Another big problem is that the image is built also in check mode, despite `--dry-run` being specified (this seems to be a bug in Compose; resp. it's not clear to me whether `--dry-run` is actually intended to be there for `docker compose build` - https://docs.docker.com/reference/cli/docker/compose/build/ doens't list it for example).

felixfontein:
I created https://github.com/docker/compose/issues/13040 for the latter. While looking at that, I also noticed that `docker compose up --dry-run` also builds the images. This is also a problem for the docker_compose_v2 module, since it's check mode isn't really check mode due to that...

felixfontein:
The Compose issue has been closed with https://github.com/docker/compose/pull/13042. Once that's released, build will simply output some generic messages which basically always say "build happened" (whether the images were actually updated or not).

Maybe we should remove check mode from this module, so that Ansible will simply skip `community.docker.docker_compose_v2_build` tasks in check mode - or alternatively always return `changed=true`. (I think the former would be more honest.)

mjbogusz:
compose [v2.39.0](https://github.com/docker/compose/releases/tag/v2.39.0) was released with the necessary fixes; maybe I'll manage to carve out some time to work on this next week, no promises though

PR #1100, opened 2025-07-25
Move EE tests to antsibull-nox

felixfontein:
##### SUMMARY
Ref: https://github.com/ansible-collections/community.dns/pull/271

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
Ah, the problem is that /var/run/docker.sock is owned by root and accessible by a group < 1000, of which user 1000 is not part. The old CI solved this by making /var/run/docker.sock world-readable...

felixfontein:
The CI failure is unrelated.

PR #1144, opened 2025-10-06
Make all doc fragments, module utils, and plugin utils private

felixfontein:
##### SUMMARY
Makes it easier to clean up and refactor...

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
various


felixfontein:
I still have no clue what's up with `plugins/module_utils/_common.py`, since the same tests (with the same Python version and same Python packages) pass locally for me. I'm going to merge this PR even if that one still fails, and try to look into this problem later... (It was also failing before this PR, so it's not related to this PR.)

PR #1192, opened 2025-11-13
docker_image(_pull), docker_container: fix compatibility with Docker 29.0.0

felixfontein:
##### SUMMARY
They're currently failing with Docker 29.0.0. Unfortunately Arch doesn't have 29.0.0 yet so I can't test this locally.

##### ISSUE TYPE
- Bugfix Pull Request
- Test Pull Request

##### COMPONENT NAME
docker_container

### DIFF
--- a/changelogs/fragments/1192-docker_container.yml
+++ b/changelogs/fragments/1192-docker_container.yml
@@ -1,4 +1,6 @@
 bugfixes:
+  - "docker_image - fix ``source=pull`` idempotency with Docker 29.0.0 (https://github.com/ansible-collections/community.docker/pull/1192)."
+  - "docker_image_pull - fix idempotency with Docker 29.0.0 (https://github.com/ansible-collections/community.docker/pull/1192)."
   - "docker_container - fix ``pull`` idempotency with Docker 29.0.0 (https://github.com/ansible-collections/community.docker/pull/1192)."
   - "docker_container - fix idempotency for IPv6 addresses with Docker 29.0.0 (https://github.com/ansible-collections/community.docker/pull/1192)."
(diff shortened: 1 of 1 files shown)


felixfontein:
It looks like Docker 29.0.0 now adds additional metadata to the pulled image, which makes the blunt check `old_tag == new_tag` (which compares the image objects, not just "tags") fail when they should be equal:
```diff
                "Metadata": {
-                   "LastTagTime": "2025-11-13T06:32:10.465631934Z"
+                   "LastTagTime": "2025-11-13T06:32:52.184742274Z"
                },
```

felixfontein:
The network idempotency problem seems to be due to different serializations of IPv6 addresses (`fdb6:feea:44b5:a58::2` vs. `fdb6:feea:44b5:0a58::2`). Looks like Docker 29.0.0 now normalizes IPv6 addresses differently (or at all).

felixfontein:
The next stop:
```
2025-11-14T19:14:53.3016972Z 11:28     "msg": "Error creating container: 400 Client Error for http+docker://localhost/v1.52/containers/create?name=ansible-docker-test-bd468d25-options: Bad Request (\"invalid JSON: invalid port '9010-9050': invalid syntax\")"
```

felixfontein:
I reported that in https://github.com/moby/moby/issues/51537.

felixfontein:
The way docker_container was handling exposed port ranges was using an undocumented feature of Docker. I've changed the behavior to work the way Docker CLI also works, with some glue that this shouldn't redeploy existing containers on Docker < 29 that were created with the same parameters.

felixfontein:
The failing tests in group 4 are unrelated (they are caused by docker_compose_v2).

PR #1208, opened 2025-11-23
Fix pull_image and push_image to handle tag@digest format correctly

### LINKED ISSUE #1207: image_label_mismatch: ignore doesn't work with tag@digest image references
##### SUMMARY
When using Docker image references that contain both a tag AND a digest (e.g., `nginx:1.0@sha256:abc123`), the `image_label_mismatch: ignore` option does not work correctly. This causes `org.opencontainers.image.*` labels from the base image to show up in diffs even when they should be ignored.

**Root Cause Analysis:**

The bug has TWO components that both need to be fixed:

1. **`parse_repository_tag()` in `plugins/module_utils/_api/utils/utils.py`**: When an image reference includes both tag and digest, the function splits by `@` first, causing the tag to be incorrectly included in the repository name portion instead of being combined with the digest.

2. **`_image_lookup()` in `plugins/module_utils/_common.py` and `_common_api.py`**: Even after `parse_repository_tag()` returns the correct `(repo, tag@digest)` tuple, the lookup function cannot match this combined format against Docker's data model. Docker stores `RepoTags` (e.g., `["nginx:1.21"]`) and `RepoDigests` (e.g., `["nginx@sha256:abc"]`) SEPARATELY - it never stores combined `tag@digest` references. The lookup function was constructing lookups like `nginx:1.21@sha256:abc` which never exist in either list.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
- `parse_repository_tag` in `plugins/module_utils/_api/utils/utils.py`
- `_image_lookup` in `plugins/module_utils/_common.py`
- `_image_lookup` in `plugins/module_utils/_common_api.py`

Affected modules/plugins that use these functions:

`docker_image`
`docker_image_pull`
`docker_image_push`
`docker_image_export`
`docker_image_tag`
`docker_image_build`
`docker_image_remove`
`docker_image_info`
`docker_swarm_service`
`docker_container` (via `_module_container/module.py`)

##### ANSIBLE VERSION
```
ansible [core 2.20.0]
  config file = /Users/paul/Source/home-server/ansible/ansible.cfg
  python version = 3.14.0
```

##### COLLECTION VERSION
```
community.docker 5.0.1
```

##### OS / ENVIRONMENT
OS: macOS 26.1 / Unraid 7.2.0
Docker version: 27.5.1
Python: 3.14.0

##### STEPS TO REPRODUCE
```yaml
- name: Deploy Docker containers
    name: "homepage"
    image: "ghcr.io/gethomepage/homepage:v1.7@sha256:b6ac42c31845ea7c862d71451c16413a6284430d8ff08e16ad791f42718a7c71"
    image_label_mismatch: ignore
    comparisons:
      labels: strict
```

Deploy the playbook and then run it with `--check --diff`.

##### EXPECTED RESULTS
No diff should show up after the deploy.

##### ACTUAL RESULTS
The `org.opencontainers.image.*` labels show up as a diff even though they should be ignored.

**Technical Explanation:**

When Ansible receives an image like `ghcr.io/gethomepage/homepage:v1.7@sha256:abc123`:

1. `parse_repository_tag()` (BEFORE fix) would return:
   - `('ghcr.io/gethomepage/homepage:v1.7', 'sha256:abc123')` - Tag incorrectly attached to repo name

2. `parse_repository_tag()` (AFTER fix) returns:
   - `('ghcr.io/gethomepage/homepage', 'v1.7@sha256:abc123')` - Tag and digest correctly combined

3. But then `_image_lookup()` (BEFORE fix) would construct:
   - `lookup = "ghcr.io/gethomepage/homepage:v1.7@sha256:abc123"` - never in RepoTags
   - `lookup_digest = "ghcr.io/gethomepage/homepage@v1.7@sha256:abc123"` - never in RepoDigests
   - Result: No match found

4. Docker actually stores:
   - RepoTags: `["ghcr.io/gethomepage/homepage:v1.7"]`
   - RepoDigests: `["ghcr.io/gethomepage/homepage@sha256:abc123"]`

5. `_image_lookup()` (AFTER fix) correctly splits the combined format:
   - Checks `"ghcr.io/gethomepage/homepage:v1.7"` in RepoTags
   - AND checks `"ghcr.io/gethomepage/homepage@sha256:abc123"` in RepoDigests
   - Result: Match found

Laptop765:
##### SUMMARY

This PR fixes a bug that caused `pull_image` and `push_image` to fail with "invalid tag format" errors when using Docker image references containing both a tag AND a digest (e.g., `nginx:1.21@sha256:abc123`).

**The Problem:**

The Docker SDK and API don't accept combined `tag@digest` in the tag parameter. When `parse_repository_tag()` returns `("nginx", "1.21@sha256:abc123")`, passing this directly to `pull()` or push API fails:

```
400 Client Error: Bad Request ("invalid tag format")
```

**The Solution:**

Added a `build_pull_arguments()` helper function that detects combined `tag@digest` format and restructures the parameters:

```python
def build_pull_arguments(name: str, tag: str) -> tuple[str, str | None]:
    """
    Build correct arguments for Docker pull/push operations.

    - Tag-only: "v1.0" -> (name, "v1.0")
    - Digest-only: "sha256:abc..." -> (name, "sha256:abc...")
    - Combined: "v1.0@sha256:abc..." -> ("name:v1.0@sha256:abc...", None)
    """
    if "@" in tag and not tag.startswith("sha256:"):
        tag_part, digest = tag.split("@", 1)
        return f"{name}:{tag_part}@{digest}", None
    return name, tag
```

Fixes #1207

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
- `build_pull_arguments` in `plugins/module_utils/_util.py` (new)
- `pull_image` in `plugins/module_utils/_common.py`
- `pull_image` in `plugins/module_utils/_common_api.py`
- `push_image` in `plugins/modules/docker_image.py`

##### ADDITIONAL INFORMATION

**Files Changed:**

1. `plugins/module_utils/_util.py` - Added `build_pull_arguments()` helper function
2. `plugins/module_utils/_common.py` - Updated SDK-based `pull_image()` to use helper
3. `plugins/module_utils/_common_api.py` - Updated API-based `pull_image()` to use helper
4. `plugins/modules/docker_image.py` - Updated `push_image()` to use helper
5. `tests/unit/plugins/module_utils/test_pull_image.py` - Comprehensive unit tests

**Test Coverage:**

18 unit tests covering:
- Tag-only formats (`nginx:1.21`, `ghcr.io/user/repo:v1.0`, `localhost:5000/app:v2.0`)
- Digest-only formats (`nginx@sha256:abc...`)
- Combined tag@digest formats (the bug case)
- Integration tests with `parse_repository_tag()` to verify end-to-end flow

**Example Fix (Pull):**

Before fix:
```python
# pull_image("portainer/portainer-ee", "2.35.0-alpine@sha256:abc...")
# -> Calls: pull("portainer/portainer-ee", tag="2.35.0-alpine@sha256:abc...")
# -> FAILS: "invalid tag format"
```

After fix:
```python
# pull_image("portainer/portainer-ee", "2.35.0-alpine@sha256:abc...")
# -> build_pull_arguments returns: ("portainer/portainer-ee:2.35.0-alpine@sha256:abc...", None)
# -> Calls: pull("portainer/portainer-ee:2.35.0-alpine@sha256:abc...")
# -> SUCCESS: Docker accepts full reference without tag parameter
```

**Example Fix (Push):**

Before fix:
```python
# push_image() calls /images/{name}/push?tag=2.35.0-alpine@sha256:abc...
# -> FAILS: "invalid tag format"
```

After fix:
```python
# push_image() calls /images/{name:tag@digest}/push (no tag param)
# -> SUCCESS: Docker accepts full reference in image name
```

All existing tests continue to pass, confirming backward compatibility.

### DIFF
--- a/plugins/module_utils/_api/utils/utils.py
+++ b/plugins/module_utils/_api/utils/utils.py
@@ -240,9 +240,23 @@ def convert_service_networks(
 
 
 def parse_repository_tag(repo_name: str) -> tuple[str, str | None]:
+    # Check for digest (@ separator) first
     parts = repo_name.rsplit("@", 1)
     if len(parts) == 2:
-        return tuple(parts)  # type: ignore
+        # We have a digest, but there might also be a tag before it
+        repo_and_tag = parts[0]
+        digest = parts[1]
+
+        # Check if there's a tag in the part before the digest
+        tag_parts = repo_and_tag.rsplit(":", 1)
+        if len(tag_parts) == 2 and "/" not in tag_parts[1]:
+            # We have both tag and digest: return repo and "tag@digest"
+            return tag_parts[0], f"{tag_parts[1]}@{digest}"
+        else:
+            # Only digest, no tag: return repo and digest
+            return repo_and_tag, digest
+
+    # No digest, check for tag only
     parts = repo_name.rsplit(":", 1)
     if len(parts) == 2 and "/" not in parts[1]:
         return tuple(parts)  # type: ignore
--- a/plugins/module_utils/_common.py
+++ b/plugins/module_utils/_common.py
@@ -27,6 +27,7 @@ from ansible_collections.community.docker.plugins.module_utils._util import (
     DOCKER_COMMON_ARGS,
     DOCKER_MUTUALLY_EXCLUSIVE,
     DOCKER_REQUIRED_TOGETHER,
+    build_pull_arguments,
     filter_images_by_tag,
     sanitize_result,
     update_tls_hostname,
@@ -572,17 +573,22 @@ class AnsibleDockerClientBase(Client):
         """
         Pull an image
         """
-        kwargs = {
-            "tag": tag,
+        kwargs: dict[str, t.Any] = {
             "stream": True,
             "decode": True,
         }
         if image_platform is not None:
             kwargs["platform"] = image_platform
+
+        # Build pull arguments - handles combined tag@digest format
+        pull_name, pull_tag = build_pull_arguments(name, tag)
+        if pull_tag is not None:
+            kwargs["tag"] = pull_tag
+
         self.log(f"Pulling image {name}:{tag}")
         old_tag = self.find_image(name, tag)
         try:
-            for line in self.pull(name, **kwargs):
+            for line in self.pull(pull_name, **kwargs):
                 self.log(line, pretty_print=True)
                 if line.get("error"):
                     if line.get("errorDetail"):
--- a/plugins/module_utils/_common.py
+++ b/plugins/module_utils/_common.py
@@ -27,6 +27,7 @@ from ansible_collections.community.docker.plugins.module_utils._util import (
     DOCKER_COMMON_ARGS,
     DOCKER_MUTUALLY_EXCLUSIVE,
     DOCKER_REQUIRED_TOGETHER,
+    filter_images_by_tag,
     sanitize_result,
     update_tls_hostname,
 )
@@ -563,18 +564,7 @@ class AnsibleDockerClientBase(Client):
             response = self.images(name=name)
         except Exception as exc:  # pylint: disable=broad-exception-caught
             self.fail(f"Error searching for image {name} - {exc}")
-        images = response
-        if tag:
-            lookup = f"{name}:{tag}"
-            lookup_digest = f"{name}@{tag}"
-            images = []
-            for image in response:
-                tags = image.get("RepoTags")
-                digests = image.get("RepoDigests")
-                if (tags and lookup in tags) or (digests and lookup_digest in digests):
-                    images = [image]
-                    break
-        return images
+        return filter_images_by_tag(name, tag, response)
 
     def pull_image(
         self, name: str, tag: str = "latest", image_platform: str | None = None
--- a/plugins/module_utils/_common_api.py
+++ b/plugins/module_utils/_common_api.py
@@ -56,6 +56,7 @@ from ansible_collections.community.docker.plugins.module_utils._util import (
     DOCKER_COMMON_ARGS,
     DOCKER_MUTUALLY_EXCLUSIVE,
     DOCKER_REQUIRED_TOGETHER,
+    build_pull_arguments,
     filter_images_by_tag,
     sanitize_result,
     update_tls_hostname,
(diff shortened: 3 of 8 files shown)


plugins/module_utils/_api/utils/utils.py
```diff
@@ -240,9 +240,23 @@ def convert_service_networks(
 
 
 def parse_repository_tag(repo_name: str) -> tuple[str, str | None]:
```
felixfontein: This function is called in a lot of places, changing its semantics can introduce a lot of subtle bugs. I don't think it's a good idea to change the behavior of this function.

I think it's better to have proper parsing for images. I've started working on that a longer time ago, but then got interrupted and haven't continued it for a longer time. The original approach in Docker SDK for Python - where this code comes from - seems to be pretty broken, and causes a lot of problems in quite a few places in this collection (and likely in other users of that code).

felixfontein (changes requested):
Thanks for your contribution!

Laptop765:
## Additional Fix: Image Lookup Idempotency

After testing the initial fix, I discovered a related bug causing **non-idempotent behavior** - `docker_container` was pulling the image on every run even when the correct image already existed locally.

### Root Cause

The `filter_images_by_tag()` function required **BOTH** RepoTags AND RepoDigests to match for combined `tag@digest` references. However, Docker does NOT store tags in `RepoTags` when pulling by digest:

```
$ docker image inspect "portainer/portainer-ee:2.35.0-alpine@sha256:7ecf2008..."
{
    "RepoTags": [],  # Empty\!
    "RepoDigests": ["portainer/portainer-ee@sha256:7ecf2008..."]
}
```

This caused `find_image()` to always return no match, triggering a pull on every ansible run.

### The Fix

When a digest is specified (combined `tag@digest` format), match by digest only since it is the authoritative identifier. The tag is informational for human readability but Docker does not preserve it.

```python
# Before (broken):
if lookup_tag in repo_tags and lookup_digest in repo_digests:
    return [image]  # Never matches - repo_tags is empty\!

# After (fixed):
if lookup_digest in repo_digests:
    return [image]  # Matches by digest
```

### Commit

Latest commit: https://github.com/Laptop765/community.docker/commit/0525a80

Tests updated to verify the correct behavior, including the critical case of empty `RepoTags`.

PR #1220, opened 2025-12-06
[stable-4] Docker Compose 5+: improve image layer event parsing (#1219)

felixfontein:
##### SUMMARY
Backport of #1219 to stable-4.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
compose_v2 module utils


felixfontein:
Looks like the JSON event parsing didn't really work with Python 2. My guess is that on the systems in CI that are still using Python 2 also use older Compose versions that do not support JSON events, so we never noticed... While the chance that a Python 2 user is using a more recent Compose version isn't exactly high, I think it's good to get this fixed.

PR #1223, opened 2025-12-14
[WIP] Use antsibull-nox also for AZP

felixfontein:
##### SUMMARY
This is an experiment. The idea is to avoid having part of the CI matrix repeated in AZP. The sole source of truth should be antsibull-nox.toml.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
It seems that AZP does not support templating a job's `displayName` with matrix values. Which, to be honest, is a real blocker to this being useful.

Maybe I should instead write a Python script to generate the AZP config...

(And avoid AZP ever again when not really needed... It's the most painful CI environment I've ever used.)

Issue #1216, opened 2025-12-02
If "docker compose images" logs to stderr, task execution fails


niels-hb:
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

community.docker.docker_compose_v2

##### ANSIBLE VERSION

```paste below
ansible [core 2.19.4]
  config file = ./ansible/ansible.cfg
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /nix/store/90jiynl0ah5z92p41xg4n2vbkdqayfci-python3.13-ansible-core-2.19.4/lib/python3.13/site-packages/ansible
  ansible collection location = ~/.ansible/collections:/usr/share/ansible/collections
  executable location = /nix/store/90jiynl0ah5z92p41xg4n2vbkdqayfci-python3.13-ansible-core-2.19.4/bin/ansible
  python version = 3.13.9 (main, Oct 14 2025, 13:52:31) [Clang 21.1.2 ] (/nix/store/hv7qldswrqvg1sjys1lmgjg875hkffng-python3-3.13.9/bin/python3.13)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

##### COLLECTION VERSION

```paste below
Collection       Version
---------------- -------
community.docker 5.0.3
```

##### CONFIGURATION

```paste below
CONFIG_FILE() = ./ansible/ansible.cfg
DEFAULT_HOST_LIST(./ansible/ansible.cfg) = ['./ansible/inventory.yaml']
DEFAULT_ROLES_PATH(./ansible/ansible.cfg) = ['./ansible/roles']
EDITOR(env: EDITOR) = nvim

GALAXY_SERVERS:
```

##### OS / ENVIRONMENT

Remote host: Rocky Linux 10.0 (Red Quartz) - x86_64
Ansible host: macOS 26.1 - Apple M3

##### STEPS TO REPRODUCE

```yaml
- name: Run docker-compose.yaml
  community.docker.docker_compose_v2:
    project_src: /opt/vault_agent
    state: "present"
```

```yaml
services:
  vault-agent:
    image: hashicorp/vault:1.21
    restart: always
```

##### EXPECTED RESULTS

`call_cli_json` should not fail if stdout doesn't contain data.

##### ACTUAL RESULTS

```
TASK [vault_agent : Run docker-compose.yaml] ***********************************
[...]

PR #1258, opened 2026-05-13
bugfix: surface CLI rc/stderr in call_cli_json instead of opaque JSONDecodeError

mdean20:
## Summary

When the docker CLI exits non-zero with empty or non-JSON stdout, the
real cause -- the daemon's error message on stderr -- is hidden behind
an opaque `Expecting value: line 1 column 1 (char 0)` `JSONDecodeError`
emitted by `json.loads(stdout)` inside `call_cli_json` /
`call_cli_json_stream` in `plugins/module_utils/_common_cli.py`.

This is the same class of problem already discussed in
#1216 and partially mitigated in #1221 (which started appending
stderr to the error message). The leading message a user sees is still
"Error while parsing JSON output of ...: Expecting value: line 1 column
1 (char 0)", which sends operators chasing a phantom JSON bug instead
of the real CLI failure that lives in stderr.

### Symptom seen in the wild

`community.docker.docker_compose_v2` running against a host whose
compose project referenced a stale `Config.Image` SHA produced:

```
docker compose images --format json
  rc:     1
  stdout: ""
  stderr: "Error response from daemon: No such image: sha256:c02a671fd71657..."
```

The module then ran `json.loads("")` and raised
`Expecting value: line 1 column 1 (char 0)`, completely hiding the
clear "No such image" diagnostic from the daemon. Multiple rounds of
debugging were required before the real cause was identified.

### Fix

In both `call_cli_json` and `call_cli_json_stream`, when JSON parsing
fails *and* the CLI also exited with `rc != 0`, raise an error that
leads with the CLI's exit code and stderr instead of the JSON parse
error. A small helper `_fail_with_cli_error` centralises the message
format. Machine-readable `cmd` / `rc` / `stdout` / `stderr` fields are
populated the same way as the existing error path.

Behavior is unchanged when:
- `stdout` parses successfully (the regular happy path), or
- `stdout` parses successfully but `rc != 0` -- e.g. `docker image
  inspect <missing>` returning `rc=1` with stdout `[]`. The `find_image`
  / `find_image_by_id` call sites that rely on this continue to work.

The fix only changes the *message text and contents* on the error path
that today produces the misleading `JSONDecodeError`.

### Changelog fragment

`changelogs/fragments/fix-call-cli-json-rc-check.yml`:

```yaml
bugfixes:
  - "CLI-based modules - when the docker CLI exits non-zero and produces no parseable JSON on stdout, surface the CLI's own ``stderr`` and exit code instead of the misleading ``Expecting value: line 1 column 1 (char 0)`` ``JSONDecodeError``. ..."
```

## Test plan

Added `tests/unit/plugins/module_utils/test__common_cli.py` with the
following cases:

- [x] `call_cli_json`: success with valid JSON returns parsed data.
- [x] `call_cli_json`: `rc=1` with valid JSON stdout `[]` (the `docker image inspect <missing>` shape) still returns; does **not** short-circuit -- preserves `find_image` semantics.
- [x] `call_cli_json`: `rc=1` with empty stdout + stderr like `No such image: sha256:...` now fails with a message containing `rc=1` and the stderr text, **not** `Expecting value` and **not** `Error while parsing JSON output`. Machine-readable `rc`/`stderr`/`stdout` fields are populated.
- [x] `call_cli_json`: `rc=0` with unparseable stdout keeps the existing `Error while parsing JSON output` message (the JSON parse error is the relevant one here).
- [x] `call_cli_json_stream`: success returns parsed lines.
- [x] `call_cli_json_stream`: `rc=2` with a malformed JSON line + `connection refused` stderr surfaces the CLI's `rc=2` / stderr instead of the JSON parse error.
- [x] `call_cli_json_stream`: `rc=0` with a malformed JSON line keeps the existing JSON parsing error message.
- [x] `call_cli_json`: `warn_on_stderr=True` still calls `self.warn(stderr)` on the success path.

Tests are pure unit tests against a minimal `AnsibleDockerClientBase`
subclass that overrides `call_cli` / `fail` / `warn`; no docker daemon
is required and no new dependencies are introduced.

Refs https://github.com/ansible-collections/community.docker/issues/1216

### DIFF
--- a/changelogs/fragments/fix-call-cli-json-rc-check.yml
+++ b/changelogs/fragments/fix-call-cli-json-rc-check.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "CLI-based modules - when the docker CLI exits non-zero and produces no parseable JSON on stdout, surface the CLI's own ``stderr`` and exit code instead of the misleading ``Expecting value: line 1 column 1 (char 0)`` ``JSONDecodeError``. This addresses cases such as a stale ``Config.Image`` SHA causing ``docker compose images --format json`` to fail with ``Error response from daemon: No such image: ...`` on stderr and empty stdout (https://github.com/ansible-collections/community.docker/issues/1216)."
--- a/plugins/module_utils/_common_cli.py
+++ b/plugins/module_utils/_common_cli.py
@@ -178,6 +178,30 @@ class AnsibleDockerClientBase:
     ) -> tuple[int, bytes, bytes]:
         pass
 
+    def _fail_with_cli_error(
+        self,
+        args: t.Sequence[str],
+        rc: int,
+        stdout: bytes,
+        stderr: bytes,
+    ) -> t.NoReturn:
+        """Fail with a clear message that leads with the CLI's own error output
+        and exit code. Used when stdout could not be parsed as JSON *and* the
+        CLI exited non-zero -- in that situation the underlying CLI error
+        (which lives in stderr) is the real cause and is far more informative
+        than the downstream JSON parsing failure.
+        """
+        cmd = self._compose_cmd_str(args)
+        self.fail(
+            f"Error while executing {cmd}: command exited with rc={rc}.\n"
+            f"Error output:\n{to_text(stderr)}\n"
+            f"Standard output:\n{to_text(stdout)}",
+            cmd=cmd,
+            rc=rc,
+            stdout=stdout,
+            stderr=stderr,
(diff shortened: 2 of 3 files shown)


felixfontein:
So the only difference to the extended message introduced in #1221 is that first stderr is shown, and then stdout, instead of the other way around?

Did the wrong order of stdout / stderr (with empty stdout) really cause that much problem?

PR #1262, opened 2026-05-30
AZP's Docker is no longer compatible with quite a few of the Docker versions needed for tests, so use GHA whenever possible

felixfontein:
##### SUMMARY
As topic says.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
Looks like this is the way...

